PR #110, opened 2021-09-19
Bolt pouch support

rbbi:
Adds support for bolt pouches

image


dillydill123:
Thanks for the PR. There are a few minor things I caught but I can handle those. Well done :)

rbbi:
This could do with some refactoring to make it work better with Rune Pouch. A lot of their code can be shared.

PR #113, opened 2021-09-23
Refactor BoltPouch and RunePouch panels into common superclass

rbbi:


dillydill123:
Great work. I think in the future I would like to create a better system in the plugin to manage these "ammunition" panels. But I'm not sure how many more I would add.. so maybe it's not worth the effort. Thanks!

PR #115, opened 2021-10-03
Fix unwanted removal of bank tab separator

Twiglet1022:
The new option to remove the bank tab separator is also applying when you're not filtering the bank and just trying to view the main all items tab. This small tweak fixes that.


dillydill123:
Nice find, thank you!

PR #128, opened 2022-04-07
add ability to configure the color of setup names

### LINKED ISSUE #123: add the ability to color the names of your setups
please, thanks!

call-me-maple:
Close #123

Hey, I added the ability to configure the color of setup names. Really not sure how this should look from a UI perspective and would love some input. Currently, the label is placed next to the name on the StandardPanel and uses the highlight color icon as a placeholder.

image image

Also, I wasn't sure if the color should be used in the nameInput label in the StandardPanel or the title in the PluginPanel.

### DIFF
--- a/src/main/java/inventorysetups/InventorySetup.java
+++ b/src/main/java/inventorysetups/InventorySetup.java
@@ -67,6 +67,10 @@ public class InventorySetup
 
 	@Getter
 	@Setter
+	private Color displayColor;
+
+	@Getter
+	@Setter
 	private boolean filterBank;
 
 	@Getter
--- a/src/main/java/inventorysetups/InventorySetupsConfig.java
+++ b/src/main/java/inventorysetups/InventorySetupsConfig.java
@@ -153,7 +153,7 @@ public interface InventorySetupsConfig extends Config
 	)
 	default Color displayColor()
 	{
-		return JagexColors.MENU_TARGET;
+		return null;
 	}
 
 	@ConfigItem(
--- a/src/main/java/inventorysetups/InventorySetupsConfig.java
+++ b/src/main/java/inventorysetups/InventorySetupsConfig.java
@@ -31,6 +31,7 @@ import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
 import net.runelite.client.config.ConfigSection;
 import net.runelite.client.config.Keybind;
+import net.runelite.client.ui.JagexColors;
 
 import static inventorysetups.InventorySetupsPlugin.CONFIG_KEY_COMPACT_MODE;
 import static inventorysetups.InventorySetupsPlugin.CONFIG_KEY_HIDE_BUTTON;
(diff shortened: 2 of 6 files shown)


dillydill123:
Hey thanks for the pull request. I'll have to play around with this a little bit, but I think the only thing I'm on the fence on is how changing the color is done. There's already such limited real estate on the on panel, and this only makes it so the maximum name length will be even shorter. Perhaps you can try other ways to prompt the color picker? Maybe right clicking the name? Or right click the paint bucket icon for an option to change color of the name? I think keeping the UI as simple and uncluttered as possible is a good move, because I often get lots of people who struggle or get overwhelmed with the UI already.

call-me-maple:
What do you think about changing the rename prompt to edit and adding there?

dillydill123:
I think that would be better. It could be changed to "edit" and a recolor option?

call-me-maple:
Thoughts on one of these?

bUS0bTfqiU
R7gerB2Tzq

dillydill123:
Either one looks fine. If you are going for the paint bucket make sure to add some space between the left border and bucket itself. Much better overall though!

dillydill123:
Are you happy with your changes?

call-me-maple:
Yep! 😊

dillydill123:
So should the color also get updated on the panel? I feel like this was the intention of the original issue

dillydill123:
So I tested it and everything appears to work fine. I added some code to also change the color of the setup name on the panel. But I'm not sure if it looks great. At the same time, I think people will be confused when the color doesn't change. So I would recommend also changing the `nameInput` in both `InventorySetupsCompactPanel` and `InventorySetupsStandardPanel`.

Something like this: `nameInput.getTextField().setForeground(invSetup.getDisplayColor());`

If you feel that this should be toggleable then you can add a config option.

call-me-maple:
Yeah, I agree that it looks weird. I found that when using a darker color for the `nameInput` the contrast makes it harder to see. I wouldn't want people to be discourage from using a color they like in the right-click menu because they don't like how it looks in the side panel. Currently, the default `displayColor` is the orange `MENU_TARGET` so we would also be changing all the `nameInputs` to that as well. Do you think there is a better way to display the current color, maybe an underline or something?  If we do change the `nameInput` color as well I think a config option would be great. 

Also should the `title` in `InventorySetupsPluginPanel` of an active setup be changed too?

call-me-maple:
This changes the `NAME_BOTTOM_BORDER` to use the setup's `displayColor` and increases the thickness by a pixel.
image

This just adds a MatteBorder to the `nameInput` in `InventorySetupsCompactPanel`
image

I like the look of this compact panel for sure.

dillydill123:
That looks great and is a good middle ground. I'll accept an extra pixel width. One question is what will the default be? Should be just white? Will that make the show worn items menu also white?

call-me-maple:
Currently, the default is orange. If we changed the default to white then yes all the worn item menus would change to white as too.

Example of default white:
image

dillydill123:
So the default of the of the colors on the panels should probably be the dark gray, but the show worn items should probably be the original target color? And only then would you change it if a different color was selected? What do you think?

call-me-maple:
That works for me.

dillydill123:
Not sure how the color is being stored but you could use null as an indicator that no color has been selected.

call-me-maple:
Made the discussed changes 😊

dillydill123:
I finally got around to this and tested it. I'll be making some additional changes on top of it (mainly for removing the color as I did not see a way to do so).

PR #129, opened 2022-04-13
update for runelite 1.8.17

majiru:
With the next runelite release the varbit api will change slightly. The following two commits detail the change:
https://github.com/runelite/runelite/commit/d129c49923beaf79ba3e4ccf10c0f097435a9cfa
https://github.com/runelite/runelite/commit/2bc2e867bdc1b07be38f486193faeadfa780845f

This updates a bit of the code to work with the newer api.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ repositories {
 }
 
 
-def runeLiteVersion = '1.8.8'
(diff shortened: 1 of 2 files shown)


dillydill123:
Thanks for catching this. Looks like I was already using it in some spots but not in others

PR #137, opened 2022-06-15
Update InventorySetupsVariationMapping.java

Bmanger:
Add support for ToB ornament kits in fuzzy mapping


Bmanger:
Hey,

I've added the ToB ornament kits to the variation mapping. Can make sharing inventory setups with people with/without ornament kits a little easier. I've only had a cursory look at the project, so I might have missed a few things.

dillydill123:
I don't have these items but the code looks good to me. Just to verify you tested and it's working appropriately?

PR #171, opened 2022-11-10
Add support for locked divine rune pouch

palleiko:


dillydill123:
@palleiko have you been able to test this? I don't have a locked divine rune pouch. I'm fairly confident it will work but always a good idea to do a real in game test.

brianbis:
can confirm this id works as expected

Issue #235, opened 2023-12-28
Request: Search item in panel to find setups it is in


Intecpsp:
If there is a way to do this already, I have not found it. But I'd love to be able to search in the panel for an item and it filters the list of setups based on if that item is in them or not.

Use case: I am migrating from bank tags and layouts to this plugin and since I'm getting rid of the tags, I won't be able to see if I can get rid of an item as easily (if it has tags or not). Being able to search for it would tell me if I can sell the item, or if it is needed. (Keeping up tags and setups is a bear, and I would prefer not to do it)

dillydill123:
@rbbi you are working on some implementation of this? I think the best way would be to have a tag in the search bar like item:<name> and it will find each setup which has the item

rbbi:
> @rbbi you are working on some implementation of this? I think the best way would be to have a tag in the search bar like item: and it will find each setup which has the item

Sounds good, created a PR for this #238

dillydill123:
Done, in 1.18

PR #238, opened 2024-02-15
Search for inventory setup via setup name or contents

rbbi:
As discussed in #235, search using the "item:" prefix to search for specific inventory/equipment/bolt pouch/rune pouch items by name.

image

### DIFF
--- a/src/main/java/inventorysetups/InventorySetupsPlugin.java
+++ b/src/main/java/inventorysetups/InventorySetupsPlugin.java
@@ -41,14 +41,7 @@ import java.lang.reflect.Type;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.Set;
+import java.util.*;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
@@ -1036,10 +1029,29 @@ public class InventorySetupsPlugin extends Plugin
 
 	public List<InventorySetup> filterSetups(String textToFilter)
 	{
-		final String textToFilterLower = textToFilter.toLowerCase();
-		return inventorySetups.stream()
-			.filter(i -> i.getName().toLowerCase().contains(textToFilterLower))
-			.collect(Collectors.toList());
+    return inventorySetups.stream().filter(i -> setupContains(i, textToFilter.toLowerCase())).collect(Collectors.toList());
+	}
+
+	private static boolean setupContains(InventorySetup i, String textToFilterLower)
+	{
+		if (textToFilterLower.startsWith("item:") && textToFilterLower.length() >= 6)
+		{
+			String itemName = textToFilterLower.substring("item:".length());
+			// Find setups containing the given item name
+			return containsItem(i.getInventory(), itemName) || containsItem(i.getEquipment(), itemName)
+					|| containsItem(i.getBoltPouch(), itemName) || containsItem(i.getRune_pouch(), itemName);
+		}
+		// Find setups containing the given setup name (default behaviour)
+		return i.getName().toLowerCase().contains(textToFilterLower);
+	}
+
+	private static boolean containsItem(List<InventorySetupsItem> inventorySetupItems, String textToFilterLower) {
+		if (inventorySetupItems == null) {
+			return false;
+		}
+		return inventorySetupItems.stream().filter(Objects::nonNull)
+        .map(item -> item.getName().toLowerCase())
+        .anyMatch(itemName -> itemName.contains(textToFilterLower));
 	}
 
 	public void doBankSearch(final InventorySetupsFilteringModeID filteringModeID)
@@ -2446,7 +2458,7 @@ public class InventorySetupsPlugin extends Plugin
 		}
 		else if (runePouchTypeOldItem != InventorySetupsRunePouchType.NONE)
 		{
-			// if the old item is a rune pouch, need to update it to null 
+			// if the old item is a rune pouch, need to update it to null
 			inventorySetup.updateRunePouch(null);
 		}
 
--- a/src/main/java/inventorysetups/InventorySetupsPlugin.java
+++ b/src/main/java/inventorysetups/InventorySetupsPlugin.java
@@ -1039,22 +1039,22 @@ public class InventorySetupsPlugin extends Plugin
 	public List<InventorySetup> filterSetups(String textToFilter)
 	{
 		return inventorySetups.stream()
-			.filter(inventorySetup -> setupContains(inventorySetup, textToFilter.toLowerCase()))
+			.filter(inventorySetup -> shouldDisplaySetup(inventorySetup, textToFilter.trim().toLowerCase()))
 			.collect(Collectors.toList());
 	}
 
-	private static boolean setupContains(InventorySetup inventorySetup, String textToFilterLower)
+	private static boolean shouldDisplaySetup(InventorySetup inventorySetup, String trimmedTextToFilterLower)
 	{
-		if (textToFilterLower.startsWith(ITEM_SEARCH_TAG) && textToFilterLower.length() > ITEM_SEARCH_TAG.length())
+		if (trimmedTextToFilterLower.startsWith(ITEM_SEARCH_TAG) && trimmedTextToFilterLower.length() > ITEM_SEARCH_TAG.length())
 		{
-			String itemName = textToFilterLower.trim().substring(ITEM_SEARCH_TAG.length());
+			String itemName = trimmedTextToFilterLower.substring(ITEM_SEARCH_TAG.length()).trim();
 			// Find setups containing the given item name
 			return containerContainsItemByName(inventorySetup.getInventory(), itemName) || containerContainsItemByName(inventorySetup.getEquipment(), itemName)
 				|| containerContainsItemByName(inventorySetup.getBoltPouch(), itemName) || containerContainsItemByName(inventorySetup.getRune_pouch(), itemName)
 				|| containerContainsItemByName(inventorySetup.getAdditionalFilteredItems().values(), itemName);
 		}
 		// Find setups containing the given setup name (default behaviour)
-		return inventorySetup.getName().toLowerCase().contains(textToFilterLower);
+		return inventorySetup.getName().toLowerCase().contains(trimmedTextToFilterLower);
 	}
 
 	private static boolean containerContainsItemByName(Collection<InventorySetupsItem> itemsInContainer, String textToFilterLower)
--- a/src/main/java/inventorysetups/InventorySetupsPlugin.java
+++ b/src/main/java/inventorysetups/InventorySetupsPlugin.java
@@ -1050,8 +1050,8 @@ public class InventorySetupsPlugin extends Plugin
 			String itemName = trimmedTextToFilterLower.substring(ITEM_SEARCH_TAG.length()).trim();
 			// Find setups containing the given item name
 			return containerContainsItemByName(inventorySetup.getInventory(), itemName) || containerContainsItemByName(inventorySetup.getEquipment(), itemName)
-				|| containerContainsItemByName(inventorySetup.getBoltPouch(), itemName) || containerContainsItemByName(inventorySetup.getRune_pouch(), itemName)
-				|| containerContainsItemByName(inventorySetup.getAdditionalFilteredItems().values(), itemName);
+				|| containerContainsItemByName(inventorySetup.getRune_pouch(), itemName) || containerContainsItemByName(inventorySetup.getAdditionalFilteredItems().values(), itemName)
+				|| containerContainsItemByName(inventorySetup.getBoltPouch(), itemName);
 		}
 		// Find setups containing the given setup name (default behaviour)
 		return inventorySetup.getName().toLowerCase().contains(trimmedTextToFilterLower);
--- a/src/main/java/inventorysetups/InventorySetupsPlugin.java
+++ b/src/main/java/inventorysetups/InventorySetupsPlugin.java
@@ -41,7 +41,15 @@ import java.lang.reflect.Type;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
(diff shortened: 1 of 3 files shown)


src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -41,14 +41,7 @@
...
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.Set;
+import java.util.*;
```
dillydill123, line 44: no `.*` here, include everything separately

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1036,10 +1029,29 @@ private boolean isNewPositionInvalid(int oldPosition, int newPosition, int size)
 
 	public List<InventorySetup> filterSetups(String textToFilter)
 	{
-		final String textToFilterLower = textToFilter.toLowerCase();
-		return inventorySetups.stream()
-			.filter(i -> i.getName().toLowerCase().contains(textToFilterLower))
-			.collect(Collectors.toList());
+    return inventorySetups.stream().filter(i -> setupContains(i, textToFilter.toLowerCase())).collect(Collectors.toList());
```
dillydill123, line 1032: fix spacing and add line breaks like it was before

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1036,10 +1029,29 @@ private boolean isNewPositionInvalid(int oldPosition, int newPosition, int size)
...
+	}
+
+	private static boolean setupContains(InventorySetup i, String textToFilterLower)
+	{
+		if (textToFilterLower.startsWith("item:") && textToFilterLower.length() >= 6)
+		{
+			String itemName = textToFilterLower.substring("item:".length());
```
dillydill123, line 1039: trim the leading and trailing whitespace in case someone adds spaces before or after the item

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1036,10 +1029,29 @@ private boolean isNewPositionInvalid(int oldPosition, int newPosition, int size)
...
 	public List<InventorySetup> filterSetups(String textToFilter)
 	{
-		final String textToFilterLower = textToFilter.toLowerCase();
-		return inventorySetups.stream()
-			.filter(i -> i.getName().toLowerCase().contains(textToFilterLower))
-			.collect(Collectors.toList());
+    return inventorySetups.stream().filter(i -> setupContains(i, textToFilter.toLowerCase())).collect(Collectors.toList());
+	}
+
+	private static boolean setupContains(InventorySetup i, String textToFilterLower)
```
dillydill123, line 1035: use final for args, also `i` is not a good name. Please use `inventorySetup` for the `InventorySetup` object

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1036,10 +1029,29 @@ private boolean isNewPositionInvalid(int oldPosition, int newPosition, int size)
...
+	{
+		if (textToFilterLower.startsWith("item:") && textToFilterLower.length() >= 6)
```
dillydill123, line 1037: Use a constant like `public static final string ITEM_SEARCH_TAG = "item:"`. Then you can also do > size instead of having just 6.

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1036,10 +1029,29 @@ private boolean isNewPositionInvalid(int oldPosition, int newPosition, int size)
...
+		{
+			String itemName = textToFilterLower.substring("item:".length());
+			// Find setups containing the given item name
+			return containsItem(i.getInventory(), itemName) || containsItem(i.getEquipment(), itemName)
```
dillydill123, line 1041: This is not checking the additional filtered items. Not sure if you want it to. It's unfortunately not a List type (which I regret a lot), so it would require a special check

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1036,10 +1029,29 @@ private boolean isNewPositionInvalid(int oldPosition, int newPosition, int size)
...
+		// Find setups containing the given setup name (default behaviour)
+		return i.getName().toLowerCase().contains(textToFilterLower);
+	}
+
+	private static boolean containsItem(List<InventorySetupsItem> inventorySetupItems, String textToFilterLower) {
+		if (inventorySetupItems == null) {
+			return false;
+		}
+		return inventorySetupItems.stream().filter(Objects::nonNull)
+        .map(item -> item.getName().toLowerCase())
```
dillydill123, line 1053: Spacing

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1036,10 +1029,29 @@ private boolean isNewPositionInvalid(int oldPosition, int newPosition, int size)
...
+			return containsItem(i.getInventory(), itemName) || containsItem(i.getEquipment(), itemName)
+					|| containsItem(i.getBoltPouch(), itemName) || containsItem(i.getRune_pouch(), itemName);
+		}
+		// Find setups containing the given setup name (default behaviour)
+		return i.getName().toLowerCase().contains(textToFilterLower);
+	}
+
+	private static boolean containsItem(List<InventorySetupsItem> inventorySetupItems, String textToFilterLower) {
+		if (inventorySetupItems == null) {
+			return false;
```
dillydill123, line 1050: spacing

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1036,10 +1029,29 @@ private boolean isNewPositionInvalid(int oldPosition, int newPosition, int size)
...
+			String itemName = textToFilterLower.substring("item:".length());
+			// Find setups containing the given item name
+			return containsItem(i.getInventory(), itemName) || containsItem(i.getEquipment(), itemName)
+					|| containsItem(i.getBoltPouch(), itemName) || containsItem(i.getRune_pouch(), itemName);
+		}
+		// Find setups containing the given setup name (default behaviour)
+		return i.getName().toLowerCase().contains(textToFilterLower);
+	}
+
+	private static boolean containsItem(List<InventorySetupsItem> inventorySetupItems, String textToFilterLower) {
```
dillydill123, line 1048: instead of `inventorySetupItems` name it `itemsInContainer`

dillydill123, line 1048: instead of `containsItem`, maybe `containerContainsItemByName`. I would want to use `containerContainsItem` but there's a similar function for ID comparison

src/main/java/inventorysetups/InventorySetupsPlugin.java
dillydill123, line 1035: instead of `setupContains`, how about `shouldFilterSetup`?

rbbi: I prefer `setupContains` because we're filtering it in both cases, just in two different ways depending on the search text.

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1036,10 +1029,29 @@ private boolean isNewPositionInvalid(int oldPosition, int newPosition, int size)
...
+		if (inventorySetupItems == null) {
+			return false;
+		}
+		return inventorySetupItems.stream().filter(Objects::nonNull)
```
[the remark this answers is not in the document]
rbbi: Seems to be okay without the filter now, will remove

src/main/java/inventorysetups/InventorySetupsPlugin.java
rbbi: Good catch, sure, I'll add this.

rbbi:
Do we want to automatically reperform the current search whenever an item is removed/added from a setup? Otherwise it becomes out-of-date.

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1036,12 +1038,36 @@ private boolean isNewPositionInvalid(int oldPosition, int newPosition, int size)
...
-			.filter(i -> i.getName().toLowerCase().contains(textToFilterLower))
+			.filter(inventorySetup -> setupContains(inventorySetup, textToFilter.toLowerCase()))
 			.collect(Collectors.toList());
 	}
 
+	private static boolean setupContains(InventorySetup inventorySetup, String textToFilterLower)
+	{
+		if (textToFilterLower.startsWith(ITEM_SEARCH_TAG) && textToFilterLower.length() > ITEM_SEARCH_TAG.length())
+		{
+			String itemName = textToFilterLower.trim().substring(ITEM_SEARCH_TAG.length());
```
dillydill123, line 1050: The `trim` before here is good but you also want to trim after. I think this still fails if someone does something like 

`   item: hammer  `. You are trimming to `item: hammer` but then fetch the string ` hammer` which has a space at the beginning.

dillydill123:
> Do we want to automatically reperform the current search whenever an item is removed/added from a setup? Otherwise it becomes out-of-date.

I don't think so. What are the ways a user can update the setup without going into the setup itself? 
   - Adding an additional filtered item
That's all I can think of. Every other situation would cause them to go into the setup, and then back out which would cause a reload. I don't think it's worth redrawing the panel

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1036,12 +1038,36 @@ private boolean isNewPositionInvalid(int oldPosition, int newPosition, int size)
...
+			// Find setups containing the given item name
+			return containerContainsItemByName(inventorySetup.getInventory(), itemName) || containerContainsItemByName(inventorySetup.getEquipment(), itemName)
+				|| containerContainsItemByName(inventorySetup.getBoltPouch(), itemName) || containerContainsItemByName(inventorySetup.getRune_pouch(), itemName)
+				|| containerContainsItemByName(inventorySetup.getAdditionalFilteredItems().values(), itemName);
```
dillydill123, line 1054: Nice use of collections here. clever!

src/main/java/inventorysetups/InventorySetupsPlugin.java
dillydill123: `setupContains` just sounds like an incomplete sentence to me. Either way this function is testing if we should filter the current setup? Maybe `shouldDisplayCurrentSetup`?

src/main/java/inventorysetups/InventorySetupsPlugin.java
dillydill123: I didn't test this but this is what I think it does, lmk if my understanding is wrong.

src/main/java/inventorysetups/InventorySetupsPlugin.java
rbbi: Yep, you're right. Okay, `shouldDisplaySetup` 👍🏼

src/main/java/inventorysetups/InventorySetupsPlugin.java
rbbi: Good catch.

rbbi:
> > Do we want to automatically reperform the current search whenever an item is removed/added from a setup? Otherwise it becomes out-of-date.
> 
> I don't think so. What are the ways a user can update the setup without going into the setup itself?
> 
> * Adding an additional filtered item
>   That's all I can think of. Every other situation would cause them to go into the setup, and then back out which would cause a reload. I don't think it's worth redrawing the panel

1. Search for an item using `item:X`
2. Open a setup containing the item
3. Remove that item from the setup
4. Click the "Return to setups" button - the search is now no longer up to date as that item has been removed

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1036,12 +1038,36 @@ private boolean isNewPositionInvalid(int oldPosition, int newPosition, int size)
...
 			.collect(Collectors.toList());
 	}
 
+	private static boolean shouldDisplaySetup(InventorySetup inventorySetup, String trimmedTextToFilterLower)
+	{
+		if (trimmedTextToFilterLower.startsWith(ITEM_SEARCH_TAG) && trimmedTextToFilterLower.length() > ITEM_SEARCH_TAG.length())
+		{
+			String itemName = trimmedTextToFilterLower.substring(ITEM_SEARCH_TAG.length()).trim();
+			// Find setups containing the given item name
+			return containerContainsItemByName(inventorySetup.getInventory(), itemName) || containerContainsItemByName(inventorySetup.getEquipment(), itemName)
```
dillydill123, line 1052: the order here is good, just check bolt pouch last since it's least likely to have the item. very small optimization. 

I think this order makes the most sense for short circuiting:
inventory, equipment, rune pouch, filtered items, bolt pouch.

rbbi: Sure

rbbi:
> > > > Do we want to automatically reperform the current search whenever an item is removed/added from a setup? Otherwise it becomes out-of-date.
> > > 
> > > 
> > > I don't think so. What are the ways a user can update the setup without going into the setup itself?
> > > 
> > > * Adding an additional filtered item
> > >   That's all I can think of. Every other situation would cause them to go into the setup, and then back out which would cause a reload. I don't think it's worth redrawing the panel
> > 
> > 
> > 
> > 1. Search for an item using `item:X`
> > 2. Open a setup containing the item
> > 3. Remove that item from the setup
> > 4. Click the "Return to setups" button - the search is now no longer up to date as that item has been removed
> 
> I see, I don't call `redrawOverviewPanel` in `returnToOverviewPanel`. Try adding this in, ideally before we try to set the scrollbar position. Hopefully this doesn't have any flickering affect. If there is, maybe we need to change the order of calls in `returnToOverviewPanel`. And performance should be ok I think.

Will take a look when I get a moment

dillydill123:
Everything looks good besides the redraw part. Let me know how that behaves and I can check it out after March 2nd (out of town)

rbbi:
Hi, calling `redrawOverviewPanel` instead of `returnToOverviewPanel` in the mouse listener causes a slight flickering for me when I return to setups. Moving the `returnToOverviewPanel` call to after the filtering call within `redrawOverviewPanel` seems to fix the flickering for me.

dillydill123:
> Hi, calling `redrawOverviewPanel` instead of `returnToOverviewPanel` in the mouse listener causes a slight flickering for me when I return to setups. Moving the `returnToOverviewPanel` call to after the filtering call within `redrawOverviewPanel` seems to fix the flickering for me.

So far I've noticed that this kills the saved scrollbar position. I'm testing it to see how we can preserve it so hitting the back arrow doesn't cause it to go to the top

dillydill123:
> > Hi, calling `redrawOverviewPanel` instead of `returnToOverviewPanel` in the mouse listener causes a slight flickering for me when I return to setups. Moving the `returnToOverviewPanel` call to after the filtering call within `redrawOverviewPanel` seems to fix the flickering for me.
> 
> So far I've noticed that this kills the saved scrollbar position. I'm testing it to see how we can preserve it so hitting the back arrow doesn't cause it to go to the top

I've swapped your lines in `redrawOverviewPanel`. Laying out the setups first must be done otherwise trying to set a long scrollbar position doesn't make sense since no setups are there and the window is very short. Given I can just add this commit after I will do so. Nice work

PR #241, opened 2024-03-02
Use separate config keys per InventorySetup

### LINKED ISSUE #206: Inventory Setups Not Saving
This issue started a week or two ago where whenever I add a new inventory setup or edit an old one, those changes are erased when I close the Runelite client and re-open it. So if I add a new setup and then close and re-open the client, that setup will have disappeared. If I edit any older setups that I made in the past and then close the client and re-open the client, those edits will have been erased and the setup will appear as it was a week or 2 ago (whenever this issue started)

My best guess for when this issue started is around 5/14/2023, as that is when I first made edits to a setup that I realized (a few days later) did not save. All the edits/new setups I had made since then were reverted, and this issue persists today. The only change/action I did around 5/14 that I can think of which maybe caused this was switching the plugin view from "standard mode" to "icon mode."

Otherwise, I use the inventory setups plugin pretty heavily, with 84 setups saved for almost every boss in the game and some skilling methods/mini-games. I also type out "summary guides" in the notes sections for tips and tricks and other things I've learned for future reference, so many of those inventory setups have large walls of text in their notes. Is it possible that my inventory setups plugin is "full" due to all that data, and that's why changes keep reverting?

Other Notes:
- I have not and am not signed into the same account on multiple Runelite clients.
- Uninstalling/re-installing Runelite does not resolve the issue, nor does logging in/out of Runelite profile.
- I'm on a Macbook.
- Any settings I change in my other Runelite plugins save whenever I close the client, so this issue seems to be unique to the inventory setups plugin.

Any help would be much appreciated.

poi56iop:
Also handles migration.

Fixes #206.

Let me know if there's anything I should fix or change; thanks!

### DIFF
--- a/src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
+++ b/src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
@@ -49,6 +49,7 @@ public class InventorySetupsPersistentDataManager
 	public static final String CONFIG_KEY_SETUPS = "setups";
 	public static final String CONFIG_KEY_SETUPS_V2 = "setupsV2";
 	public static final String CONFIG_KEY_SETUPS_V3_PREFIX = "setupsV3_";
+	public static final String CONFIG_KEY_SETUPS_ORDER_V3 = "setupsOrderV3_";
 	public static final String CONFIG_KEY_SECTIONS = "sections";
 
 	@Inject
@@ -92,9 +93,6 @@ public class InventorySetupsPersistentDataManager
 		sections.addAll(loadData(CONFIG_KEY_SECTIONS, sectionType));
 		for (final InventorySetupsSection section : sections)
 		{
-			final String newName = InventorySetupUtilities.findNewName(section.getName(), cache.getSectionNames().keySet());
-			section.setName(newName);
-
 			// Remove any duplicates that exist
 			List<String> uniqueSetups = section.getSetups().stream().distinct().collect(Collectors.toList());
 			section.setSetups(uniqueSetups);
@@ -118,9 +116,11 @@ public class InventorySetupsPersistentDataManager
 			final List<String> oldSetupKeys = configManager.getConfigurationKeys(wholePrefix);
 			Set<String> oldSetupHashes = oldSetupKeys.stream().map(key -> key.substring(wholePrefix.length())).collect(Collectors.toSet());
 
+			final List<String> setupsOrder = new ArrayList<>();
 			for (final InventorySetup setup : inventorySetups)
 			{
 				final String hash = hashFunction.hashUnencodedChars(setup.getName()).toString();
+				setupsOrder.add(hash);
 				oldSetupHashes.remove(hash);
 				final String data = gson.toJson(InventorySetupSerializable.convertFromInventorySetup(setup));
 				configManager.setConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_V3_PREFIX + hash, data);
@@ -131,6 +131,9 @@ public class InventorySetupsPersistentDataManager
 				// Any hashes still in the oldSetupHashes set were for setups that were either renamed (and saved to a new hash above) or deleted.
 				configManager.unsetConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_V3_PREFIX + removedSetupHash);
 			}
+
+			final String setupsOrderJson = gson.toJson(setupsOrder);
+			configManager.setConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_ORDER_V3, setupsOrderJson);
 		}
 
 		if (updateSections)
@@ -185,6 +188,19 @@ public class InventorySetupsPersistentDataManager
 		return isList;
 	}
 
+	private InventorySetup loadV3Setup(String configKey) {
+		final String storedData = configManager.getConfiguration(CONFIG_GROUP, configKey);
+		try
+		{
+			return InventorySetupSerializable.convertToInventorySetup(gson.fromJson(storedData, InventorySetupSerializable.class));
+		}
+		catch (Exception e)
+		{
+			log.error(String.format("Exception occurred while loading %s", configKey), e);
+			throw e;
+		}
+	}
+
 	private List<InventorySetup> loadV3Setups() {
 		final String wholePrefix = ConfigManager.getWholeKey(CONFIG_GROUP, null, CONFIG_KEY_SETUPS_V3_PREFIX);
 		final List<String> loadedSetupWholeKeys = configManager.getConfigurationKeys(wholePrefix);
@@ -192,20 +208,25 @@ public class InventorySetupsPersistentDataManager
 			key -> key.substring(wholePrefix.length()-CONFIG_KEY_SETUPS_V3_PREFIX.length())
 		).collect(Collectors.toSet());
 
+		Type setupsOrderType = new TypeToken<ArrayList<String>>()
+		{
+
+		}.getType();
+		final String setupsOrderJson = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_ORDER_V3);
+		final List<String> setupsOrder = gson.fromJson(setupsOrderJson, setupsOrderType);
+
 		List<InventorySetup> loadedSetups = new ArrayList<>();
+		for (final String configHash : setupsOrder)
+		{
+			final String configKey = CONFIG_KEY_SETUPS_V3_PREFIX + configHash;
+			if (loadedSetupKeys.remove(configKey)) { // Handles if hash is present only in configOrder.
+				loadedSetups.add(loadV3Setup(configKey));
+			}
+		}
 		for (final String configKey : loadedSetupKeys)
 		{
-			final String storedData = configManager.getConfiguration(CONFIG_GROUP, configKey);
-
-			try
-			{
-				loadedSetups.add(InventorySetupSerializable.convertToInventorySetup(gson.fromJson(storedData, InventorySetupSerializable.class)));
-			}
-			catch (Exception e)
-			{
-				log.error(String.format("Exception occurred while loading %s", configKey), e);
-				throw e;
-			}
+			// Load any remaining setups not present in setupsOrder. Useful if updateConfig crashes midway.
+			loadedSetups.add(loadV3Setup(configKey));
 		}
 		return loadedSetups;
 	}
@@ -232,8 +253,6 @@ public class InventorySetupsPersistentDataManager
 				setup.updateAdditionalItems(new HashMap<>());
 			}
 
-			final String newName = InventorySetupUtilities.findNewName(setup.getName(), cache.getInventorySetupNames().keySet());
-			setup.setName(newName);
 			cache.addSetup(setup);
 
 			// add Item names
@@ -276,17 +295,14 @@ public class InventorySetupsPersistentDataManager
 			updateConfig(true, false);
 			inventorySetups.clear();
 			configManager.setConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V2, "True");
+			configManager.setConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V3, "True");
 		}
 
-		hasMigratedToV2 = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V2);
-		if (!Strings.isNullOrEmpty(hasMigratedToV2))
+		String oldV1Data = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS);
+		if (oldV1Data != null)
 		{
-			String oldData = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS);
-			if (oldData != null)
-			{
-				log.info("Removing old v1 data key");
-				configManager.unsetConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS);
-			}
+			log.info("Removing old v1 data key");
+			configManager.unsetConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS);
 		}
 
 		String hasMigratedToV3 = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V3);
@@ -300,15 +316,11 @@ public class InventorySetupsPersistentDataManager
 		}
 
 		// TODO Don't unset configuration until new version is stable
-//		hasMigratedToV3 = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V3);
-//		if (!Strings.isNullOrEmpty(hasMigratedToV3))
+//		String oldV2Data = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_V2);
+//		if (oldV2Data != null)
 //		{
-//			String oldData = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_V2);
-//			if (oldData != null)
-//			{
-//				log.info("Removing old v2 data key");
-//				configManager.unsetConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_V2);
-//			}
+//			log.info("Removing old v2 data key");
+//			configManager.unsetConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_V2);
 //		}
 	}
 
--- a/src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
+++ b/src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
@@ -166,7 +166,8 @@ public class InventorySetupsPersistentDataManager
 		}
 	}
 
-	private List<InventorySetup> loadV1Setups() {
+	private List<InventorySetup> loadV1Setups()
+	{
 		Type setupTypeV1 = new TypeToken<ArrayList<InventorySetup>>()
 		{
 
@@ -174,7 +175,8 @@ public class InventorySetupsPersistentDataManager
 		return loadData(CONFIG_KEY_SETUPS, setupTypeV1);
 	}
 
-	private List<InventorySetup> loadV2Setups() {
+	private List<InventorySetup> loadV2Setups()
+	{
 		Type setupTypeV2 = new TypeToken<ArrayList<InventorySetupSerializable>>()
 		{
 
@@ -188,7 +190,8 @@ public class InventorySetupsPersistentDataManager
 		return isList;
 	}
 
-	private InventorySetup loadV3Setup(String configKey) {
+	private InventorySetup loadV3Setup(String configKey)
+	{
 		final String storedData = configManager.getConfiguration(CONFIG_GROUP, configKey);
 		try
 		{
@@ -201,11 +204,12 @@ public class InventorySetupsPersistentDataManager
 		}
 	}
 
-	private List<InventorySetup> loadV3Setups() {
+	private List<InventorySetup> loadV3Setups()
+	{
 		final String wholePrefix = ConfigManager.getWholeKey(CONFIG_GROUP, null, CONFIG_KEY_SETUPS_V3_PREFIX);
 		final List<String> loadedSetupWholeKeys = configManager.getConfigurationKeys(wholePrefix);
 		Set<String> loadedSetupKeys = loadedSetupWholeKeys.stream().map(
-			key -> key.substring(wholePrefix.length()-CONFIG_KEY_SETUPS_V3_PREFIX.length())
+			key -> key.substring(wholePrefix.length() - CONFIG_KEY_SETUPS_V3_PREFIX.length())
 		).collect(Collectors.toSet());
 
 		Type setupsOrderType = new TypeToken<ArrayList<String>>()
@@ -213,13 +217,18 @@ public class InventorySetupsPersistentDataManager
 
 		}.getType();
 		final String setupsOrderJson = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_ORDER_V3);
-		final List<String> setupsOrder = gson.fromJson(setupsOrderJson, setupsOrderType);
+		List<String> setupsOrder = gson.fromJson(setupsOrderJson, setupsOrderType);
+		if (setupsOrder == null)
+		{
+			setupsOrder = new ArrayList<>();
+		}
 
 		List<InventorySetup> loadedSetups = new ArrayList<>();
 		for (final String configHash : setupsOrder)
 		{
 			final String configKey = CONFIG_KEY_SETUPS_V3_PREFIX + configHash;
-			if (loadedSetupKeys.remove(configKey)) { // Handles if hash is present only in configOrder.
+			if (loadedSetupKeys.remove(configKey))
+			{ // Handles if hash is present only in configOrder.
 				final InventorySetup setup = loadV3Setup(configKey);
 				loadedSetups.add(setup);
 			}
@@ -227,7 +236,7 @@ public class InventorySetupsPersistentDataManager
 		for (final String configKey : loadedSetupKeys)
 		{
 			// Load any remaining setups not present in setupsOrder. Useful if updateConfig crashes midway.
-			log.info(String.format("Loading setup that was missing from Order key: %s"), configKey);
+			log.info("Loading setup that was missing from Order key: " + configKey);
 			final InventorySetup setup = loadV3Setup(configKey);
 			loadedSetups.add(setup);
 		}
@@ -293,7 +302,7 @@ public class InventorySetupsPersistentDataManager
 		String hasMigratedToV2 = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V2);
 		if (Strings.isNullOrEmpty(hasMigratedToV2))
 		{
-			log.info("Migrating data to V2");
+			log.info("Migrating data from V1 to V3");
 			inventorySetups.addAll(loadV1Setups());
 			updateConfig(true, false);
 			inventorySetups.clear();
@@ -311,7 +320,7 @@ public class InventorySetupsPersistentDataManager
 		String hasMigratedToV3 = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V3);
 		if (Strings.isNullOrEmpty(hasMigratedToV3))
 		{
-			log.info("Migrating data to V3");
+			log.info("Migrating data from V2 to V3");
 			inventorySetups.addAll(loadV2Setups());
 			updateConfig(true, false);
 			inventorySetups.clear();
--- a/src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
+++ b/src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
@@ -220,13 +220,16 @@ public class InventorySetupsPersistentDataManager
 		{
 			final String configKey = CONFIG_KEY_SETUPS_V3_PREFIX + configHash;
 			if (loadedSetupKeys.remove(configKey)) { // Handles if hash is present only in configOrder.
-				loadedSetups.add(loadV3Setup(configKey));
+				final InventorySetup setup = loadV3Setup(configKey);
+				loadedSetups.add(setup);
 			}
 		}
 		for (final String configKey : loadedSetupKeys)
 		{
 			// Load any remaining setups not present in setupsOrder. Useful if updateConfig crashes midway.
-			loadedSetups.add(loadV3Setup(configKey));
+			log.info(String.format("Loading setup that was missing from Order key: %s"), configKey);
+			final InventorySetup setup = loadV3Setup(configKey);
+			loadedSetups.add(setup);
 		}
 		return loadedSetups;
 	}
--- a/src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
+++ b/src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
@@ -1,5 +1,7 @@
 package inventorysetups;
 
+import com.google.common.hash.HashFunction;
+import com.google.common.hash.Hashing;
 import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
@@ -19,6 +21,7 @@ import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import static inventorysetups.InventorySetupsPlugin.CONFIG_GROUP;
@@ -37,9 +40,15 @@ public class InventorySetupsPersistentDataManager
 	private final List<InventorySetup> inventorySetups;
 	private final List<InventorySetupsSection> sections;
 
+
+	// Fast non-cryptographic hash with relatively short output. Could use murmur3_32_fixed instead; but being extra cautious of collisions.
+	public static final HashFunction hashFunction = Hashing.murmur3_128();
+
 	public static final String CONFIG_KEY_SETUPS_MIGRATED_V2 = "migratedV2";
+	public static final String CONFIG_KEY_SETUPS_MIGRATED_V3 = "migratedV3";
 	public static final String CONFIG_KEY_SETUPS = "setups";
 	public static final String CONFIG_KEY_SETUPS_V2 = "setupsV2";
+	public static final String CONFIG_KEY_SETUPS_V3_PREFIX = "setupsV3_";
 	public static final String CONFIG_KEY_SECTIONS = "sections";
 
 	@Inject
@@ -72,15 +81,8 @@ public class InventorySetupsPersistentDataManager
 		// Handles migration of old setup data
 		handleMigrationOfOldData();
 
-		Type setupTypeV2 = new TypeToken<ArrayList<InventorySetupSerializable>>()
-		{
-
-		}.getType();
-		List<InventorySetupSerializable> issList = new ArrayList<>(loadData(CONFIG_KEY_SETUPS_V2, setupTypeV2));
-		for (final InventorySetupSerializable iss : issList)
-		{
-			inventorySetups.add(InventorySetupSerializable.convertToInventorySetup(iss));
-		}
+		final List<InventorySetup> setupsFromConfig = loadV3Setups();
+		inventorySetups.addAll(setupsFromConfig);
 		processSetupsFromConfig();
 
 		Type sectionType = new TypeToken<ArrayList<InventorySetupsSection>>()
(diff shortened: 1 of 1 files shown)


src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -72,15 +77,7 @@ public void loadConfig()
...
-		Type setupTypeV2 = new TypeToken<ArrayList<InventorySetupSerializable>>()
-		{
-
-		}.getType();
-		List<InventorySetupSerializable> issList = new ArrayList<>(loadData(CONFIG_KEY_SETUPS_V2, setupTypeV2));
-		for (final InventorySetupSerializable iss : issList)
-		{
-			inventorySetups.add(InventorySetupSerializable.convertToInventorySetup(iss));
-		}
+		inventorySetups.addAll(loadV3Setups());
```
dillydill123, line 80: I prefer we pull out `loadV3Setups()` into a var first, helps with debugging

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -108,14 +105,27 @@ public void updateConfig(boolean updateSetups, boolean updateSections)
 	{
 		if (updateSetups)
 		{
-			List<InventorySetupSerializable> issList = new ArrayList<>();
+			// Rather than escaping the name to pick a config key, instead use hash of the name so that changes to the escaping function or supported characters in the name string don't affect config keys. Also avoids issues from overly long names and prevents other plugins from using the config key as the source of truth for the name; encouraging the use of the name key within each setup's json.
```
dillydill123, line 108: Put this comment on separate lines. Also, the second half is a bit unclear to me, can you elaborate?

poi56iop: New phrasing should hopefully be clearer. I wanted to avoid representing the name twice on disk; in a real user-specified format (saved in the json under name key) and a potentially unstable escaped string; that would change if we change the escaping function. This way, plugins like bank tab layouts would definitely grab name from the json instead of using the escaped name.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -108,14 +106,30 @@ public void updateConfig(boolean updateSetups, boolean updateSections)
 	{
 		if (updateSetups)
 		{
-			List<InventorySetupSerializable> issList = new ArrayList<>();
+			// Rather than escaping the name to pick a config key, instead use hash of the name. Benefits:
+			//   Changes to the escaping function or supported characters in the name string don't affect config keys.
+			//   Standardizes length; preventing overflow issues from extremely long names
+			//   Keeps json as the single source of truth for names; other plugins won't try to grab names from config keys.
+			final String wholePrefix = ConfigManager.getWholeKey(CONFIG_GROUP, null, CONFIG_KEY_SETUPS_V3_PREFIX);
```
dillydill123, line 113: can you explain what the null is? I see it's profile. Is that rs profile as in rs account or runelite profile?

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -108,14 +106,30 @@ public void updateConfig(boolean updateSetups, boolean updateSections)
...
+			// Rather than escaping the name to pick a config key, instead use hash of the name. Benefits:
+			//   Changes to the escaping function or supported characters in the name string don't affect config keys.
+			//   Standardizes length; preventing overflow issues from extremely long names
+			//   Keeps json as the single source of truth for names; other plugins won't try to grab names from config keys.
+			final String wholePrefix = ConfigManager.getWholeKey(CONFIG_GROUP, null, CONFIG_KEY_SETUPS_V3_PREFIX);
+			final List<String> oldSetupKeys = configManager.getConfigurationKeys(wholePrefix);
+			Set<String> oldSetupHashes = oldSetupKeys.stream().map(k -> k.substring(wholePrefix.length())).collect(Collectors.toSet());
+
+			// Fast non-cryptographic hash with relatively short output. Could use murmur3_32_fixed instead; but being extra cautious of collisions.
+			final HashFunction hf = Hashing.murmur3_128();
```
dillydill123, line 118: Should we instead make this a member variable of the data manager class? If this is not a singleton then we are creating a brand new hashing function each time. Seems like we can just create one and keep it forever.

dillydill123: also, a name better than `hf`. `hashFunction` is fine

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -108,14 +106,30 @@ public void updateConfig(boolean updateSetups, boolean updateSections)
...
 		if (updateSetups)
 		{
-			List<InventorySetupSerializable> issList = new ArrayList<>();
+			// Rather than escaping the name to pick a config key, instead use hash of the name. Benefits:
+			//   Changes to the escaping function or supported characters in the name string don't affect config keys.
+			//   Standardizes length; preventing overflow issues from extremely long names
+			//   Keeps json as the single source of truth for names; other plugins won't try to grab names from config keys.
+			final String wholePrefix = ConfigManager.getWholeKey(CONFIG_GROUP, null, CONFIG_KEY_SETUPS_V3_PREFIX);
+			final List<String> oldSetupKeys = configManager.getConfigurationKeys(wholePrefix);
+			Set<String> oldSetupHashes = oldSetupKeys.stream().map(k -> k.substring(wholePrefix.length())).collect(Collectors.toSet());
```
dillydill123, line 115: k => key

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -148,6 +162,35 @@ private <T> List<T> loadData(final String configKey, Type type)
...
+		List<InventorySetup> isList = new ArrayList<>();
+		for (final InventorySetupSerializable iss : issList)
+		{
+			isList.add(InventorySetupSerializable.convertToInventorySetup(iss));
+		}
+		return isList;
+	}
+
+	private List<InventorySetup> loadV3Setups() {
+		// Keeping this as a distinct function will streamline future migrations.
```
dillydill123, line 180: unnecessary comment

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: It's a singleton; we could also move it into the class if you want.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: It looks like there is now support for saving directly to specific configuration profiles; but the old pattern of not specifying a profile remains intact. In discord conversation with Adam, he made no suggestion of switching to the new pattern, so it seems that what this plugin currently does (not specify profile) will remain supported in the future.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
dillydill123: Yeah why not, pretty minor but helps clear up that it's only being created once even if that is the case already

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: > I believe it is rsProfile; there are some `getWholeKey(groupName, rsProfile, key)` calls in the same file. that `getWholeKey` is defined in.
> 
> It looks like there is support for saving directly to specific rsProfile configurations; but the current pattern of not specifying an rsProfile remains intact. In discord conversation with Adam, he made no suggestion of switching to the new pattern, so it seems that what this plugin currently does (not specify rsProfile) will remain supported in the future.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: changing in another place too

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -148,6 +162,35 @@ private <T> List<T> loadData(final String configKey, Type type)
...
+		final String wholePrefix = ConfigManager.getWholeKey(CONFIG_GROUP, null, CONFIG_KEY_SETUPS_V3_PREFIX);
+		final List<String> loadedSetupWholeKeys = configManager.getConfigurationKeys(wholePrefix);
+		Set<String> loadedSetupKeys = loadedSetupWholeKeys.stream().map(k -> k.substring(wholePrefix.length()-CONFIG_KEY_SETUPS_V3_PREFIX.length())).collect(Collectors.toSet());
```
dillydill123: k => key, also can make this one multiple lines

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: removed

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done; set it as `public static`; which is definitely debatable though. lmk if you want me to change that.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -148,6 +165,34 @@ private <T> List<T> loadData(final String configKey, Type type)
...
+	private List<InventorySetup> loadV3Setups() {
+		final String wholePrefix = ConfigManager.getWholeKey(CONFIG_GROUP, null, CONFIG_KEY_SETUPS_V3_PREFIX);
+		final List<String> loadedSetupWholeKeys = configManager.getConfigurationKeys(wholePrefix);
+		Set<String> loadedSetupKeys = loadedSetupWholeKeys.stream().map(key -> key.substring(wholePrefix.length()-CONFIG_KEY_SETUPS_V3_PREFIX.length())).collect(Collectors.toSet());
+
+		List<InventorySetup> loadedSetups = new ArrayList<>();
+		for (final String configKey : loadedSetupKeys)
+		{
+			final String storedData = configManager.getConfiguration(CONFIG_GROUP, configKey);
+			loadedSetups.add(InventorySetupSerializable.convertToInventorySetup(gson.fromJson(storedData, InventorySetupSerializable.class)));
```
dillydill123, line 193: So far the only thing missing here that was different in V2 was some sort of error handling. If this line fails, the whole plugin will fail. I'm not sure if we actually need error handling in a practical sense, but at least if there is some bug or data corruption or something like that, we can get a log message when the `gson.fromJson` call fails

Long story short: I think some try catch should be added here in case the gson fails

poi56iop: The old handling worried me because on failure it basically wiped everything. Should the new one be the same way?

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -231,6 +276,28 @@ private void handleMigrationOfOldData()
 //				log.info("Removing old data key");
 //				configManager.unsetConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS);
 //			}
+//		}
```
dillydill123, line 281: I think we can enable this unset of the original V1 data (been over 1.5 years since that update it looks like). Additionally, line 259 to 266 can be turned into a `loadV1Setups` function

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
dillydill123: hmm yeah I see your point. Well in this case we can only wipe one inventory setup and try to continue. Or we let the plugin crash? I'm not really keen on that one, but at least if it does crash it would cause people to make an issue pretty quickly lol. Maybe do a try catch and give a better error message if the gson does fail then rethrow it?

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -231,6 +278,28 @@ private void handleMigrationOfOldData()
 //				log.info("Removing old data key");
```
dillydill123, line 278: old v1 data key

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: I left out the last three lines to match the others (not running `updateConfig`, `inventorySetups.clear`, and `configManager.setConfiguration` in `loadV1Setups`.) Let me know if I misunderstood.

Also uncommented the V1-removing-code.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done; also added for the post-v3 migration

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -231,6 +278,28 @@ private void handleMigrationOfOldData()
...
+		}
+
+		// TODO Don't unset configuration until new version is stable
+//		hasMigratedToV3 = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V3);
+//		if (!Strings.isNullOrEmpty(hasMigratedToV3))
+//		{
+//			String oldData = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_V2);
+//			if (oldData != null)
+//			{
+//				log.info("Removing old data key");
```
dillydill123: old v2 data key

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: I'd honestly rather the plugin halt execution, (as it would if gson threw an exception,) because the next time data is saved to disk, every setup is wiped.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: Also I would (perhaps naively) assume that an exception from gson would show in runelite logs even without us doing something special; because it would cause the whole plugin to fail. Only better thing I can think of would be to halt execution and have the plugin panel tell the user their data is corrupt, and give them a button to export/wipe it. IMO sounds like a new PR though.

poi56iop: I wrote those last two comments without seeing your reply; forgot I had to refresh to get GitHub to update.

I think that catch, log, rethrow would be good for now. That there aren't a bunch of "all my setups have been wiped" issues means gson errors are rare. If people start reporting that they encounter the error; then creating a pop up or panel could make sense.

poi56iop: done; added `log.error(String.format("Exception occurred while loading %s", configKey), e);` before rethrowing.

poi56iop:
I believe every comment has been handled.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -108,14 +110,29 @@ public void updateConfig(boolean updateSetups, boolean updateSections)
...
-			List<InventorySetupSerializable> issList = new ArrayList<>();
+			// Rather than escaping the name to pick a config key, instead use hash of the name. Benefits:
+			//   Changes to the escaping function or supported characters in the name string don't affect config keys.
+			//   Standardizes length; preventing overflow issues from extremely long names
+			//   Keeps json as the single source of truth for names; other plugins won't try to grab names from config keys.
+			final String wholePrefix = ConfigManager.getWholeKey(CONFIG_GROUP, null, CONFIG_KEY_SETUPS_V3_PREFIX);
+			final List<String> oldSetupKeys = configManager.getConfigurationKeys(wholePrefix);
+			Set<String> oldSetupHashes = oldSetupKeys.stream().map(key -> key.substring(wholePrefix.length())).collect(Collectors.toSet());
+
+			// Fast non-cryptographic hash with relatively short output. Could use murmur3_32_fixed instead; but being extra cautious of collisions.
```
dillydill123, line 121: Old comment

dillydill123:
Gonna do some testing on my side. did you test importing?

dillydill123:
Also, could you fixup/squash your commits into one?

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: Good catch; removed.

poi56iop:
> Gonna do some testing on my side. did you test importing?

I did not; I didn't think that workflow would be affected.

> Also, could you fixup/squash your commits into one?

Done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -148,6 +163,53 @@ private <T> List<T> loadData(final String configKey, Type type)
...
+			}
+			catch (Exception e)
+			{
+				log.error(String.format("Exception occurred while loading %s", configKey), e);
+				throw e;
+			}
+		}
+		return loadedSetups;
+	}
+
```
dillydill123, line 212: Line 173/235 is now irrelevant. If someone has previous duplicate name setups, they will get overwritten in `updateConfig` and this code won't do anything. We could fix this by checking for duplicate names during V1/V2 migration, but This is probably not necessary since it's been a very long time since I fixed the duplicate names.

```
final String newName = InventorySetupUtilities.findNewName(setup.getName(), cache.getInventorySetupNames().keySet());
setup.setName(newName);
```

You can delete these lines. I tested it by creating a duplicate entry in the config (with duplicate name), and all that happens is one of the duplicates gets overwritten.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -210,26 +272,42 @@ private void handleMigrationOfOldData()
 		if (Strings.isNullOrEmpty(hasMigratedToV2))
 		{
 			log.info("Migrating data to V2");
-			// Perform migration of old data
-			Type setupType = new TypeToken<ArrayList<InventorySetup>>()
+			inventorySetups.addAll(loadV1Setups());
+			updateConfig(true, false);
+			inventorySetups.clear();
+			configManager.setConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V2, "True");
```
dillydill123, line 278: This here is incorrect. I tested a situation where someone still had V1 data and never migrated to V2. All the setups get wiped. If a migration to V2 never happened, then we go straight to V3, and I think we can set both MIGRATED_V2 and MIGRATED_V3 to true:

```
configManager.setConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V2, "True");
configManager.setConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V3, "True");
```

Pretty niche situation but might help someone out there with V1 data who comes back to play again.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -210,26 +272,42 @@ private void handleMigrationOfOldData()
...
 			log.info("Migrating data to V2");
-			// Perform migration of old data
-			Type setupType = new TypeToken<ArrayList<InventorySetup>>()
+			inventorySetups.addAll(loadV1Setups());
+			updateConfig(true, false);
+			inventorySetups.clear();
+			configManager.setConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V2, "True");
+		}
+
+		hasMigratedToV2 = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V2);
```
dillydill123, line 281: I think the wrapping if statement and `hasMigratedToV2` assignment can be removed. Based on the order in the function, the data is already migrated when we reach here. So there is no need to check.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -210,26 +272,42 @@ private void handleMigrationOfOldData()
...
 			updateConfig(true, false);
 			inventorySetups.clear();
-			configManager.setConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V2, "True");
+			configManager.setConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V3, "True");
 		}
 
 		// TODO Don't unset configuration until new version is stable
-//		hasMigratedToV2 = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V2);
-//		if (!Strings.isNullOrEmpty(hasMigratedToV2))
+//		hasMigratedToV3 = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V3);
```
dillydill123, line 303: you can remove the assignmetn and wrapping if statement here too, (but still leave it commneted out

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -210,26 +272,42 @@ private void handleMigrationOfOldData()
...
+//		if (!Strings.isNullOrEmpty(hasMigratedToV3))
 //		{
-//			String oldData = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS);
+//			String oldData = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_V2);
```
dillydill123, line 306: `oldV2Data`

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -210,26 +272,42 @@ private void handleMigrationOfOldData()
...
 //			if (oldData != null)
```
dillydill123, line 307: `oldV2Data`

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -210,26 +272,42 @@ private void handleMigrationOfOldData()
...
+			inventorySetups.addAll(loadV1Setups());
+			updateConfig(true, false);
+			inventorySetups.clear();
+			configManager.setConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V2, "True");
+		}
+
+		hasMigratedToV2 = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V2);
+		if (!Strings.isNullOrEmpty(hasMigratedToV2))
+		{
+			String oldData = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS);
```
dillydill123, line 284: `oldV1Data`

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -210,26 +272,42 @@ private void handleMigrationOfOldData()
...
+			if (oldData != null)
```
dillydill123, line 285: `oldV1Data`

dillydill123:
I made some comments regarding the testing I've done. Additionally, there's one problem (side effect?) of the current implementation. Setups are not in any particular order when read. Previously, setups had an inherent order that was stable and wouldn't change unless the user changed the order specifically (right click menus).  Now setups have an order define by what order the config manager fetches the keys.

I would like to preserve the "natural" order of the setups. Previously it was just the order setups were created in. Not sure the best way to do this at the moment. I had two ideas so far

You could keep track of a list which contains the hashes of the setup names in order. Given that each hash is 32 chars, you could realistically have ~250000/32 = over ~7000 setups before we hit the config line limit, at which point I'm pretty sure there will be other problems lol. This seems like the easiest approach. First read this sorted list, then find each V3 key in order and load the setups. Since we batch process everything anyways even for one singular update, it should be fine just to recreate this list order every time and update the config value.

Another solution could be to add a underscore `_` after the hash with a number indicating the position of the setup, so like `inventorysetups.setupsV3.hash_0` would indicate position 0. This seems more difficult. Each time a setup moves or is created or is deleted, some of the positions need to be updated.

Other things I've tested:

- Migration from V1 to V3 (broken atm, see my comments)
- Moving setups (Doesn't do anything anymore)
- Creating and Deleting Setups (working)
- Editing Setups with slot, search, and removing items. (working)
- Creating sections, adding setups, moving setups within sections (working)
- Importing and mass importing setups (working)
- Loading setups from show worn items (working)

poi56iop:
Very good points, and great catch on the ordering; I didn't notice the stable ordering as a feature of the original implementation.

To avoid frequent mass-config-key-rewrites, something like the first solution would be better. I'll need to investigate/understand how ordering within different sections is handled.
I'd be tempted to use the fix as an opportunity to handle the unique IDing if it can solve any problems here. If we were only concerned about sorting by creation order, using epoch of creation/import time would be a reasonable solution, but we also allow manual re-ordering by the user. Rather than having an ordering-preservation config key refer to setups by hash while the sections-key refers to them by name, it could be time to unify on an ID that could be used by bank tag layouts etc. as discussed in discord.

Thinking about it more though, it'll probably be easier to keep changes isolated to serialization/deserialization in this file. I'm guessing per-section ordering will be unaffected, so your proposal should solve things nicely and without much extra code.

I'll try to work on it soon.

dillydill123:
> Very good points, and great catch on the ordering; I didn't notice the stable ordering as a feature of the original implementation.
> 
> To avoid frequent mass-config-key-rewrites, something like the first solution would be better. I'll need to investigate/understand how ordering within different sections is handled. I'd be tempted to use the fix as an opportunity to handle the unique IDing if it can solve any problems here. If we were only concerned about sorting by creation order, using epoch of creation/import time would be a reasonable solution, but we also allow manual re-ordering by the user. Rather than having an ordering-preservation config key refer to setups by hash while the sections-key refers to them by name, it could be time to unify on an ID that could be used by bank tag layouts etc. as discussed in discord.
> 
> Thinking about it more though, it'll probably be easier to keep changes isolated to serialization/deserialization in this file. I'm guessing per-section ordering will be unaffected, so your proposal should solve things nicely and without much extra code.
> 
> I'll try to work on it soon.

section and section ordering should not be affected. It’s not ideal to have one organized by hash and one organized by name, but it’s totally serviceable. For bank tag layouts, all that matters is the setup name or hash, so I would avoid touching sections for now.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done; great catch

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

poi56iop:
> I would like to preserve the "natural" order of the setups.
> [...]
> * Migration from V1 to V3 (broken atm, see my comments)
> * Moving setups (Doesn't do anything anymore)

I believe these should all be fixed.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -148,6 +166,71 @@ private <T> List<T> loadData(final String configKey, Type type)
...
+		}.getType();
+		final String setupsOrderJson = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_ORDER_V3);
+		final List<String> setupsOrder = gson.fromJson(setupsOrderJson, setupsOrderType);
+
+		List<InventorySetup> loadedSetups = new ArrayList<>();
+		for (final String configHash : setupsOrder)
+		{
+			final String configKey = CONFIG_KEY_SETUPS_V3_PREFIX + configHash;
+			if (loadedSetupKeys.remove(configKey)) { // Handles if hash is present only in configOrder.
+				loadedSetups.add(loadV3Setup(configKey));
```
dillydill123, line 223: move the inner function call to the line before. Just my preference and easier to debug

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -148,6 +166,71 @@ private <T> List<T> loadData(final String configKey, Type type)
...
+			}
+		}
+		for (final String configKey : loadedSetupKeys)
+		{
+			// Load any remaining setups not present in setupsOrder. Useful if updateConfig crashes midway.
+			loadedSetups.add(loadV3Setup(configKey));
```
dillydill123, line 229: move the inner function call to the line before.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -148,6 +166,71 @@ private <T> List<T> loadData(final String configKey, Type type)
...
+		for (final String configHash : setupsOrder)
+		{
+			final String configKey = CONFIG_KEY_SETUPS_V3_PREFIX + configHash;
+			if (loadedSetupKeys.remove(configKey)) { // Handles if hash is present only in configOrder.
+				loadedSetups.add(loadV3Setup(configKey));
+			}
+		}
+		for (final String configKey : loadedSetupKeys)
+		{
+			// Load any remaining setups not present in setupsOrder. Useful if updateConfig crashes midway.
```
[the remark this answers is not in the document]
poi56iop: Correct; this'll only really come into play if updateConfig crashes midway.

Done.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

dillydill123:
gonna do some more testing and try to break it

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -210,27 +294,36 @@ private void handleMigrationOfOldData()
 		if (Strings.isNullOrEmpty(hasMigratedToV2))
 		{
 			log.info("Migrating data to V2");
```
dillydill123, line 296: `"Migrating data from V1 to V3"`

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -210,27 +294,36 @@ private void handleMigrationOfOldData()
...
+		if (oldV1Data != null)
+		{
+			log.info("Removing old v1 data key");
+			configManager.unsetConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS);
+		}
+
+		String hasMigratedToV3 = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_MIGRATED_V3);
+		if (Strings.isNullOrEmpty(hasMigratedToV3))
+		{
+			log.info("Migrating data to V3");
```
dillydill123, line 314: `"Migrating data from V2 to V3`"

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -148,6 +166,74 @@ private <T> List<T> loadData(final String configKey, Type type)
...
+			log.error(String.format("Exception occurred while loading %s", configKey), e);
+			throw e;
+		}
+	}
+
+	private List<InventorySetup> loadV3Setups() {
+		final String wholePrefix = ConfigManager.getWholeKey(CONFIG_GROUP, null, CONFIG_KEY_SETUPS_V3_PREFIX);
+		final List<String> loadedSetupWholeKeys = configManager.getConfigurationKeys(wholePrefix);
+		Set<String> loadedSetupKeys = loadedSetupWholeKeys.stream().map(
+			key -> key.substring(wholePrefix.length()-CONFIG_KEY_SETUPS_V3_PREFIX.length())
```
dillydill123, line 208: add white space around the `-`

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -148,6 +166,74 @@ private <T> List<T> loadData(final String configKey, Type type)
...
 
+	private List<InventorySetup> loadV1Setups() {
+		Type setupTypeV1 = new TypeToken<ArrayList<InventorySetup>>()
+		{
+
+		}.getType();
+		return loadData(CONFIG_KEY_SETUPS, setupTypeV1);
+	}
+
+	private List<InventorySetup> loadV2Setups() {
```
dillydill123, line 177: Check style gives me a warning about curly braces on same line. Could you move these to below the function?

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -148,6 +166,74 @@ private <T> List<T> loadData(final String configKey, Type type)
...
+
+		Type setupsOrderType = new TypeToken<ArrayList<String>>()
+		{
+
+		}.getType();
+		final String setupsOrderJson = configManager.getConfiguration(CONFIG_GROUP, CONFIG_KEY_SETUPS_ORDER_V3);
+		final List<String> setupsOrder = gson.fromJson(setupsOrderJson, setupsOrderType);
+
+		List<InventorySetup> loadedSetups = new ArrayList<>();
+		for (final String configHash : setupsOrder)
```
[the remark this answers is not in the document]
poi56iop: That would be because of the weird state of the config, but I can handle it anyway. done.

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
```diff
@@ -148,6 +166,74 @@ private <T> List<T> loadData(final String configKey, Type type)
...
+			final String configKey = CONFIG_KEY_SETUPS_V3_PREFIX + configHash;
+			if (loadedSetupKeys.remove(configKey)) { // Handles if hash is present only in configOrder.
+				final InventorySetup setup = loadV3Setup(configKey);
+				loadedSetups.add(setup);
+			}
+		}
+		for (final String configKey : loadedSetupKeys)
+		{
+			// Load any remaining setups not present in setupsOrder. Useful if updateConfig crashes midway.
+			log.info(String.format("Loading setup that was missing from Order key: %s"), configKey);
```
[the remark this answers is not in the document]
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java
poi56iop: done

dillydill123:
Looks good. I tried to break it many different ways but could not. Things start to get weird past 500 setups. menu entries in show worn items will start breaking. I'll just toss a warning about that in the readme

PR #255, opened 2024-06-12
Add config option to remember last active inventory-setup,

1Defence:
Loads on plugin/client startup, the config value is cleared if you close a setups view, prior to closing the client to prevent weird cases of a user getting thrown back into a setup when they didn't want it.

https://github.com/dillydill123/inventory-setups/assets/63735241/fde5640a-5331-449b-9b28-23bc354cfa7a

### DIFF
--- a/src/main/java/inventorysetups/InventorySetupsPlugin.java
+++ b/src/main/java/inventorysetups/InventorySetupsPlugin.java
@@ -136,6 +136,8 @@ public class InventorySetupsPlugin extends Plugin
 	public static final String CONFIG_KEY_VERSION_STR = "version";
 	public static final String CONFIG_KEY_UNASSIGNED_MAXIMIZED = "unassignedMaximized";
 	public static final String CONFIG_KEY_MANUAL_BANK_FILTER = "manualBankFilter";
+	public static final String CONFIG_KEY_REMEMBER_LAST_SETUP = "rememberLastSetup";
+	public static final String CONFIG_KEY_LAST_SETUP_VALUE = "lastSetupValue";
 	public static final String CONFIG_KEY_PERSIST_HOTKEYS = "persistHotKeysOutsideBank";
 	public static final String TUTORIAL_LINK = "https://github.com/dillydill123/inventory-setups#inventory-setups";
 	public static final String SUGGESTION_LINK = "https://github.com/dillydill123/inventory-setups/issues";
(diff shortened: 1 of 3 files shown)


1Defence:
@dillydill123 any chance something like this could be added? I know it's not a big deal when playing one account but for people who play multiple it's quite painful to go to the bank and realize you don't have it open anymore cause of a client restart especially during a death.

dillydill123:
@1Defence Thanks for the explanation. Can you explain what you mean by client restart? Or maybe describe the use case?

1Defence:
> @1Defence Thanks for the explanation. Can you explain what you mean by client restart? Or maybe describe the use case?

I'll re-explain with more detail but by "client restart" I'm referring to when you close, and re open the client.

So, I mostly use this plugin for its bank filtering functionality, say you have multiple profiles each with a setup open.
Whenever you go to the bank you'll expect it to be filtered to your setup.

**If you close the client, this is not remembered the next time you start the client.**

When starting the client your previous setup is not remembered as open and as such the bank is no longer filtered, have to open side bar to find the setup and reopen it on every account you're using.

If you use normal bank tags, you'll note this is actually a feature of them called "**Remember last tag tab**" where it'll remember it even if you close your client.

^this is explicitly the feature I was trying to implement via a config defaulted to off, as I know this is mostly an issue for people who play on more than one account.

src/main/java/inventorysetups/ui/InventorySetupsPluginPanel.java
```diff
@@ -456,6 +446,7 @@ public void mousePressed(MouseEvent e)
 				if (SwingUtilities.isLeftMouseButton(e))
 				{
 					redrawOverviewPanel(false);
+					plugin.setConfigValue(CONFIG_KEY_LAST_SETUP_VALUE, "");
```
dillydill123, line 449: Could you just set this in `shutDown()` instead so it remembers if a setup was last opened before the client was closed? Should cover all graceful paths automatically I think

src/main/java/inventorysetups/ui/InventorySetupsPluginPanel.java
```diff
@@ -777,16 +745,18 @@ public void setCurrentInventorySetup(final InventorySetup inventorySetup, boolea
...
 		if (resetScrollBar)
 		{
 			// reset scrollbar back to top
 			setScrollBarPosition(0);
 		}
 
+		plugin.setBankFilteringMode(InventorySetupsFilteringModeID.ALL);
 		plugin.doBankSearch();
 
+		plugin.setConfigValue(CONFIG_KEY_LAST_SETUP_VALUE, inventorySetup.getName());
```
dillydill123, line 759: Same thing here, I think setting this in `shutDown()` should be fine?

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -947,6 +796,97 @@ public void toggleAlphabeticalMode(InventorySetupsSortingID mode)
...
+					return false;
+			}
+
+			clientThread.invokeLater(() ->
+			{
+				dataManager.loadConfig();
+				SwingUtilities.invokeLater(() ->
+				{
+					panel.redrawOverviewPanel(true);
+					if (config.rememberLastSetup())
```
dillydill123, line 871: Maybe a private function to get the lastRememberedSetup (return null if any of these gates fail)

dillydill123 (commented):
There are merge conflicts. Rebase and I can review again

src/main/java/inventorysetups/ui/InventorySetupsPluginPanel.java
dillydill123: I think `shutDown` is triggered on any graceful exit and plugin shutdown, so might be the best option.

PR #276, opened 2024-10-06
Update README badges to api.runelite.net

TheStonedTurtle:
The website [i.pluginhub.info](https://i.pluginhub.info), which powers your README badges, is going to be shut down at the end of the month (November 1st, 2024) as you can now use RuneLite's API to generate these shields (as was done in this PR).

I highly recommend you merge this PR, or make this change yourself, before the shutdown date to avoid your badges breaking.

You should also create a PR to the hub so that your readme updates on [RuneLite.net's Plugin Hub](https://runelite.net/plugin-hub/show/inventory-setups) website after merging this PR, but that's up to you

Anyway, thanks for using my website while it was alive :heart:


dillydill123:
Thanks!

PR #351, opened 2025-07-18
Toggle fuzzy for all

### LINKED ISSUE #267: [Suggestion] (Un)fuzzy all identical items in setup
It would be nice to be able to add a shift+right-click option that would (un)fuzzy all identical items found in the current setup. Definitely helps in cases where we have the auto-fuzzy option and we want to make it so our inventory setup only contains 4 dose potions. Instead of going through and un-fuzzying all of the potions one-by-one, it would be a huge QoL to be able to quickly un-fuzzy all identical items (potions in this case).

Dragonkiller93:
Adds an option to fuzzy all, which will make all items of the same type fuzzy/nonfuzzy, but only in the one setup. Ironically, the toggle all fuzzy is not fuzzy, so it will only work for items which are exactly the same.

java_PF8CjN3krD

Wording probably needs to be changed, not sure how the menu option should be setup. 

Maybe a separate category of multiple item changers would be helpful. Similar to how shift changes all items across all setups, but only for the current setup. Could use ctrl instead of shift, and add this fuzzy toggle and a change all items option.

Fixes #267

### DIFF
--- a/src/main/java/inventorysetups/InventorySetupsPlugin.java
+++ b/src/main/java/inventorysetups/InventorySetupsPlugin.java
@@ -1679,7 +1679,7 @@ public class InventorySetupsPlugin extends Plugin
 			final List<InventorySetupsItem> container = getContainerFromSlot(slot);
 			item = container.get(slot.getIndexInSlot());
 		}
-		item.toggleIsFuzzy();
+		item.setFuzzy(!(item.isFuzzy()));
 		final int itemId = item.getId();
 		clientThread.invoke(() ->
 		{
@@ -1694,6 +1694,34 @@ public class InventorySetupsPlugin extends Plugin
 		panel.refreshCurrentSetup();
 	}
 
+	public void toggleAllFuzzyOnSlot(final InventorySetupsSlot slot) {
+		if (panel.getCurrentSelectedSetup() == null) {
+			return;
+		}
+		log.debug("Toggling all the fuzzies");
+		final List<InventorySetupsItem> container = getContainerFromSlot(slot);
+		InventorySetupsItem item = container.get(slot.getIndexInSlot());
+		boolean fuzz = item.isFuzzy();
+
+        for (final InventorySetupsItem eachItem : container) {
+            if (eachItem.getId() == getContainerFromSlot(slot).get(slot.getIndexInSlot()).getId()) {
+                eachItem.setFuzzy(!fuzz);
+            }
+        }
+
+		final int itemId = item.getId();
+		clientThread.invoke(() ->
+		{
+			if (itemId == -1)
+			{
+				return;
+			}
+			layoutUtilities.recalculateLayout(slot.getParentSetup());
+		});
+
+		dataManager.updateConfig(true, false);
+		panel.refreshCurrentSetup();
+	}
 	public void setStackCompareOnSlot(final InventorySetupsSlot slot, final InventorySetupsStackCompareID newStackCompare)
 	{
 		if (panel.getCurrentSelectedSetup() == null)
--- a/src/main/java/inventorysetups/ui/InventorySetupsAdditionalItemsPanel.java
+++ b/src/main/java/inventorysetups/ui/InventorySetupsAdditionalItemsPanel.java
@@ -90,6 +90,7 @@ public class InventorySetupsAdditionalItemsPanel extends InventorySetupsContaine
 			{
 				final InventorySetupsSlot newSlot = new InventorySetupsSlot(ColorScheme.DARKER_GRAY_COLOR, InventorySetupsSlotID.ADDITIONAL_ITEMS, i);
 				InventorySetupsSlot.addFuzzyMouseListenerToSlot(plugin, newSlot);
+				InventorySetupsSlot.addAllFuzzyMouseListenerToSlot(plugin, newSlot);
 				InventorySetupsSlot.addUpdateFromSearchMouseListenerToSlot(plugin, newSlot, false);
 				InventorySetupsSlot.addRemoveMouseListenerToSlot(plugin, newSlot);
 				additionalFilteredSlots.add(newSlot);
--- a/src/main/java/inventorysetups/ui/InventorySetupsSlot.java
+++ b/src/main/java/inventorysetups/ui/InventorySetupsSlot.java
@@ -257,7 +257,15 @@ public class InventorySetupsSlot extends JPanel
 			plugin.toggleFuzzyOnSlot(slot);
 		});
 	}
-
+	public static void addAllFuzzyMouseListenerToSlot(final InventorySetupsPlugin plugin, final InventorySetupsSlot slot)
+	{
+		JMenuItem makeAllSlotFuzzy = new JMenuItem("Toggle Fuzzy All");
+		slot.getRightClickMenu().add(makeAllSlotFuzzy);
+		makeAllSlotFuzzy.addActionListener(e ->
+		{
+			plugin.toggleAllFuzzyOnSlot(slot);
+		});
+	}
 	// adds the menu option to update set a slot to fuzzy
 	public static void addStackMouseListenerToSlot(final InventorySetupsPlugin plugin, final InventorySetupsSlot slot)
 	{
--- a/src/main/java/inventorysetups/InventorySetupsItem.java
+++ b/src/main/java/inventorysetups/InventorySetupsItem.java
@@ -46,10 +46,6 @@ public class InventorySetupsItem
 	@Setter
 	private InventorySetupsStackCompareID stackCompare;
 
-	public void toggleIsFuzzy()
-	{
-		fuzzy = !fuzzy;
-	}
 
 	public static InventorySetupsItem getDummyItem()
 	{
--- a/src/main/java/inventorysetups/ui/InventorySetupsEquipmentPanel.java
+++ b/src/main/java/inventorysetups/ui/InventorySetupsEquipmentPanel.java
@@ -60,6 +60,7 @@ public class InventorySetupsEquipmentPanel extends InventorySetupsContainerPanel
 		{
 			final InventorySetupsSlot setupSlot = new InventorySetupsSlot(ColorScheme.DARKER_GRAY_COLOR, InventorySetupsSlotID.EQUIPMENT, slot.getSlotIdx());
 			InventorySetupsSlot.addFuzzyMouseListenerToSlot(plugin, setupSlot);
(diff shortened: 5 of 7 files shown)


src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1694,6 +1694,34 @@ public void toggleFuzzyOnSlot(final InventorySetupsSlot slot)
 		panel.refreshCurrentSetup();
 	}
 
+	public void toggleAllFuzzyOnSlot(final InventorySetupsSlot slot) {
+		if (panel.getCurrentSelectedSetup() == null) {
+			return;
+		}
+		log.debug("Toggling all the fuzzies");
```
dillydill123, line 1701: Not needed

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1694,6 +1694,34 @@ public void toggleFuzzyOnSlot(final InventorySetupsSlot slot)
...
+		final List<InventorySetupsItem> container = getContainerFromSlot(slot);
+		InventorySetupsItem item = container.get(slot.getIndexInSlot());
+		boolean fuzz = item.isFuzzy();
```
dillydill123, line 1704:
```suggestion
boolean itemIsFuzzy = originalItem.isFuzzy
```

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1679,7 +1679,7 @@ public void toggleFuzzyOnSlot(final InventorySetupsSlot slot)
 			final List<InventorySetupsItem> container = getContainerFromSlot(slot);
 			item = container.get(slot.getIndexInSlot());
 		}
-		item.toggleIsFuzzy();
+		item.setFuzzy(!(item.isFuzzy()));
```
dillydill123, line 1682:
```suggestion
		item.setFuzzy(!item.isFuzzy());
```

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1694,6 +1694,34 @@ public void toggleFuzzyOnSlot(final InventorySetupsSlot slot)
 		panel.refreshCurrentSetup();
 	}
 
+	public void toggleAllFuzzyOnSlot(final InventorySetupsSlot slot) {
+		if (panel.getCurrentSelectedSetup() == null) {
+			return;
+		}
+		log.debug("Toggling all the fuzzies");
+		final List<InventorySetupsItem> container = getContainerFromSlot(slot);
+		InventorySetupsItem item = container.get(slot.getIndexInSlot());
```
dillydill123, line 1703:
```suggestion
		InventorySetupsItem originalItem = container.get(slot.getIndexInSlot());
```

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1694,6 +1694,34 @@ public void toggleFuzzyOnSlot(final InventorySetupsSlot slot)
...
+		boolean fuzz = item.isFuzzy();
+
+        for (final InventorySetupsItem eachItem : container) {
```
dillydill123, line 1706: spacing

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1694,6 +1694,34 @@ public void toggleFuzzyOnSlot(final InventorySetupsSlot slot)
...
+            if (eachItem.getId() == getContainerFromSlot(slot).get(slot.getIndexInSlot()).getId()) {
```
dillydill123, line 1707: Isn't `getContainerFromSlot(slot).get(slot.getIndexInSlot()).getId()` just the `originalItem.getId()`? No need to recompute this everytime.

src/main/java/inventorysetups/InventorySetupsPlugin.java
dillydill123, line 1706: `eachItem` -> `item`

dillydill123 (commented):
The change is fine, but I don't see the need for a separate right click menu. Why not add this login to the callback of the `Toggle Fuzzy` menu?

src/main/java/inventorysetups/ui/InventorySetupsAdditionalItemsPanel.java
```diff
@@ -90,6 +90,7 @@ public void updatePanelWithSetupInformation(InventorySetup setup)
 			{
 				final InventorySetupsSlot newSlot = new InventorySetupsSlot(ColorScheme.DARKER_GRAY_COLOR, InventorySetupsSlotID.ADDITIONAL_ITEMS, i);
 				InventorySetupsSlot.addFuzzyMouseListenerToSlot(plugin, newSlot);
+				InventorySetupsSlot.addAllFuzzyMouseListenerToSlot(plugin, newSlot);
```
dillydill123, line 93: I don't think additional filtered items would work with this at the moment since you are relying on `getContainerFromSlot`.

Unfortunately I made the additional filtered items a map instead of a list (still regret this decision), and `getContainerFromSlot` won't support it. What I've done in the past is check directly if the slot is a Additional filtered item one and behave differently based on that.

Dragonkiller93:
Lots of changes, thank you for all the help, and sorry so much was needed, still new to plugins. Should work across the board, but I did change a bit of logic that might need looking at.

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1495,7 +1495,7 @@ public void updateSlotFromSearch(final InventorySetupsSlot slot, boolean allowSt
 								panel.getCurrentSelectedSetup().getAdditionalFilteredItems();
 						if (!additionalFilteredItemsHasItem(finalId, additionalFilteredItems))
 						{
-							removeAdditionalFilteredItem(slot, additionalFilteredItems);
+							removeAdditionalFilteredItem(slot);
 							addAdditionalFilteredItem(finalId, slot.getParentSetup(), additionalFilteredItems);
 						}
 						return;
```
Dragonkiller93, line 1501: I'm not sure why this uses panel.setup.afi instead of slot.parent.afi  If needed I can modify the changes so this functions as written again.

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1708,32 +1777,42 @@ public void setStackCompareOnSlot(final InventorySetupsSlot slot, final Inventor
...
 				break;
 			}
 			j++;
 		}
+		return keyToMakeFuzzy;
+	}
+	private void removeAdditionalFilteredItem(final InventorySetupsSlot slot)
+	{
+
+		assert panel.getCurrentSelectedSetup() != null : "Current setup is null";
```
dillydill123, line 1807: I would remove this assert and if the current selected setup is null just return

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1694,6 +1684,85 @@ public void toggleFuzzyOnSlot(final InventorySetupsSlot slot)
...
 
+
+	public void toggleAllFuzzyOnSlot(final InventorySetupsSlot slot)
+	{
+		if (panel.getCurrentSelectedSetup() == null)
+		{
+			return;
+		}
+
+		final InventorySetup owner = slot.getParentSetup();
```
dillydill123, line 1695:
```suggestion
		final InventorySetup inventorySetup = slot.getParentSetup();
```

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1694,6 +1684,85 @@ public void toggleFuzzyOnSlot(final InventorySetupsSlot slot)
...
+			}
+		}
+		else{
+			originalItem = getContainerFromSlot(slot).get(slot.getIndexInSlot());
+		}
+		if (originalItem.getId() == -1)
+		{
+			return;
+		}
+		boolean fuzz = originalItem.isFuzzy();
```
dillydill123, line 1716:
```suggestion
		boolean newFuzzyValue = !originalItem.isFuzzy();
```

And use this value instead of `!fuzz`

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -1694,6 +1684,85 @@ public void toggleFuzzyOnSlot(final InventorySetupsSlot slot)
...
+		//has to be in client to use itemManager
+		clientThread.invoke(() ->
+		{
+			//Find all variations of the item, so they can also be toggled
+			int processedId = itemManager.canonicalize(originalItem.getId());
+			int baseProcessedId = InventorySetupsVariationMapping.map(processedId);
+			Collection<Integer> variations = InventorySetupsVariationMapping.getVariations(baseProcessedId);
+			variations.add(processedId);
+
+			for (final InventorySetupsItem item : owner.getInventory())
```
dillydill123, line 1727: Maybe you could use a common lambda to simplify this

```
Consumer<InventorySetupsItem> toggleFuzzyIfMatched = item ->
{
    if (variations.contains(item.getId()))
    {
        item.setFuzzy(!fuzz);
    }
};

inventorySetup.getEquipment().forEach(toggleFuzzyIfMatched);
inventorySetup.getInventory().forEach(toggleFuzzyIfMatched);
inventorySetup.getAdditionalFilteredItems().values().forEach(toggleFuzzyIfMatched);
if (inventorySetup.getQuiver() != null)
{
    inventorySetup.getQuiver().forEach(toggleFuzzyIfMatched);
}
```

src/main/java/inventorysetups/ui/InventorySetupsSlot.java
```diff
@@ -257,7 +257,15 @@ public static void addFuzzyMouseListenerToSlot(final InventorySetupsPlugin plugi
 			plugin.toggleFuzzyOnSlot(slot);
 		});
 	}
-
+	public static void addAllFuzzyMouseListenerToSlot(final InventorySetupsPlugin plugin, final InventorySetupsSlot slot)
```
dillydill123, line 260: Again, this looks fine, but I think this should just be the default. I mean that we don't need a `Toggle Fuzzy All` and `Toggle Fuzzy`. I think `Toggle Fuzzy` should just toggle all fuzzy?

dillydill123:
Looking better, more comments.

PR #356, opened 2025-08-01
Allow Bottom To Top ZigZag Layouts

edmeyer8851:
Adds an option to select Top To Bottom or Bottom To Top for ZigZag layouts -  for people with bottom to top muscle memory that want to use this plugin.

zig-zag-type-ezgif com-video-to-gif-converter(1)

### DIFF
--- a/src/main/java/inventorysetups/InventorySetupLayoutUtilities.java
+++ b/src/main/java/inventorysetups/InventorySetupLayoutUtilities.java
@@ -97,7 +97,6 @@ public class InventorySetupLayoutUtilities
 		layout.resize(newSizeGuess);
 		final HashMap<Integer, Integer> counter = new HashMap<>();
 
-
 		int nextPos = layoutZigZagContainer(setup.getEquipment(), layout, tag, addToTag, startOfEquipment, counter);
 		if (setup.getQuiver() != null && !setup.getQuiver().isEmpty())
 		{
@@ -152,8 +151,10 @@ public class InventorySetupLayoutUtilities
 		// But this is not needed, so I won't spend time over engineering this function.
 
 		int doubleRowStart = 0;
-		int nextPos = 0;
+		boolean topToBottom = config.zigZagType().equals(InventorySetupsZigZagTypeID.TOP_TO_BOTTOM);
 		final int rowSize = 8;
+
+		int nextPos = topToBottom ? 0 : 8;
 
 		for (final InventorySetupsItem item : container)
 		{
@@ -163,21 +164,43 @@ public class InventorySetupLayoutUtilities
 				continue;
 			}
 
-			if (nextPos == (rowSize * 2) - 1)
+            if (topToBottom)
 			{
-				// We hit the end of a double row, we need to start a new one.
-				doubleRowStart += 2;
-				nextPos = doubleRowStart * rowSize;
-			}
-			else if (nextPos < ((doubleRowStart * rowSize) + rowSize))
-			{
-				// We are in the top half of a double row. Go down directly one.
-				nextPos += rowSize;
-			}
+				if (nextPos == ((doubleRowStart * rowSize) + (2 * rowSize) - 1))
+				{
+					// We hit the end of a double row, we need to start a new one.
+					doubleRowStart += 2;
+					nextPos = doubleRowStart * rowSize;
+				}
+				else if (nextPos < ((doubleRowStart * rowSize) + rowSize))
+				{
+					// We are in the top half of a double row. Go down directly one.
+					nextPos += rowSize;
+				}
+				else
+				{
+					// We are in the bottom half of a double. Go back up and add one to move to the right.
+					nextPos = (nextPos - rowSize) + 1;
+				}
+            }
 			else
 			{
-				// We are in the bottom half of a double. Go back up and add one to move to the right.
-				nextPos = (nextPos - rowSize) + 1;
+				if (nextPos == ((doubleRowStart * rowSize) + rowSize - 1))
+				{
+					// We hit the end of a double row, we need to start a new one.
+					doubleRowStart += 2;
+					nextPos = (doubleRowStart * rowSize) + rowSize;
+				}
+				else if (nextPos > ((doubleRowStart * rowSize) + rowSize) - 1)
+				{
+					// We are in the bottom half of a double. Go up directly one.
+					nextPos -= rowSize;
+				}
+				else
+				{
+					// We are in the top half of a double row. Go back down and add one to move to the right.
+					nextPos = nextPos + rowSize + 1;
+				}
 			}
 		}
 
--- a/src/main/java/inventorysetups/InventorySetupsPlugin.java
+++ b/src/main/java/inventorysetups/InventorySetupsPlugin.java
@@ -145,6 +145,7 @@ public class InventorySetupsPlugin extends Plugin
 	public static final String CONFIG_KEY_PERSIST_HOTKEYS = "persistHotKeysOutsideBank";
 	public static final String CONFIG_KEY_USE_LAYOUTS = "useLayouts";
 	public static final String CONFIG_KEY_LAYOUT_DEFAULT = "defaultLayout";
+	public static final String CONFIG_KEY_ZIGZAG_TYPE = "zigZagType";
 	public static final String CONFIG_KEY_LAYOUT_DUPLICATES = "addDuplicatesInLayouts";
 	public static final String CONFIG_KEY_ENABLE_LAYOUT_WARNING = "enableLayoutWarning";
 	public static final String CONFIG_GROUP_HUB_BTL = "banktaglayouts";
--- a/src/main/java/inventorysetups/InventorySetupsZigZagTypeID.java
+++ b/src/main/java/inventorysetups/InventorySetupsZigZagTypeID.java
@@ -1,0 +1,33 @@
+package inventorysetups;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public enum InventorySetupsZigZagTypeID {
+    // Use if you want to withdraw using a top to bottom zigzag pattern
+    TOP_TO_BOTTOM("Top to Bottom", 0),
+
+    // Use if you want to withdraw using a bottom to top zigzag pattern
+    BOTTOM_TO_TOP("Bottom to Top", 1);
+
+    private final int type;
+
+    private static final List<InventorySetupsZigZagTypeID> VALUES;
+
+    static
+    {
+        VALUES = new ArrayList<>();
+        Collections.addAll(VALUES, InventorySetupsZigZagTypeID.values());
+    }
+
+    InventorySetupsZigZagTypeID(String s, int type)
+    {
+        this.type = type;
+    }
+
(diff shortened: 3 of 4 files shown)


src/main/java/inventorysetups/InventorySetupsZigZagTypeID.java
```diff
@@ -0,0 +1,33 @@
+package inventorysetups;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public enum InventorySetupsZigZagTypeID {
+    //use if you want to withdraw using a top to bottom zigzag pattern
+    Top_To_Bottom(0),
```
dillydill123, line 9:
```suggestion
TOP_TO_BOTTOM("Top to Bottom", 0)
```

src/main/java/inventorysetups/InventorySetupsZigZagTypeID.java
```diff
@@ -0,0 +1,33 @@
...
+
+    // use if you want to withdraw using a bottom to top zigzag pattern
+    Bottom_To_Top(1);
```
dillydill123, line 12:
```suggestion
BOTTOM_TO_TOP("BOTTOM_TO_TOP", 1)
```

src/main/java/inventorysetups/InventorySetupsZigZagTypeID.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public enum InventorySetupsZigZagTypeID {
+    //use if you want to withdraw using a top to bottom zigzag pattern
+    Top_To_Bottom(0),
+
+    // use if you want to withdraw using a bottom to top zigzag pattern
```
dillydill123, line 11: capitalization
```
// Use if you want to withdraw using a bottom to top zigzag pattern
```

src/main/java/inventorysetups/InventorySetupLayoutUtilities.java
```diff
@@ -67,17 +67,19 @@ public Layout getSetupLayout(final InventorySetup setup)
 	public Layout createSetupLayout(final InventorySetup setup)
 	{
 		InventorySetupLayoutType type = config.defaultLayout();
-		return createSetupLayout(setup, type, true);
+		InventorySetupsZigZagTypeID zigZagTypeID = config.zigZagType();
+		return createSetupLayout(setup, type,zigZagTypeID, true);
 	}
 
-	public Layout createSetupLayout(final InventorySetup setup, InventorySetupLayoutType type, final boolean addToTag)
+	public Layout createSetupLayout(final InventorySetup setup, InventorySetupLayoutType type, InventorySetupsZigZagTypeID zigZagTypeID, final boolean addToTag)
```
dillydill123, line 74: the zigZagID is not needed here. You are pulling it from the config in `getZigZagLayout`.

src/main/java/inventorysetups/InventorySetupLayoutUtilities.java
```diff
@@ -67,17 +67,19 @@ public Layout getSetupLayout(final InventorySetup setup)
...
 	{
 		if (type.equals(InventorySetupLayoutType.PRESET))
 		{
 			return getPresetLayout(setup, addToTag);
 		}
 		else
 		{
+
```
dillydill123, line 82: Remove this space

src/main/java/inventorysetups/InventorySetupLayoutUtilities.java
```diff
@@ -184,6 +201,42 @@ else if (nextPos < ((doubleRowStart * rowSize) + rowSize))
 		return nextPos;
 	}
 
+	private int layoutZigZagContainerBottomToTop(final List<InventorySetupsItem> container, final Layout layout, final String tag, boolean addToTag, final int start, final Map<Integer, Integer> counter)
```
dillydill123, line 204: Any way you could combine this with the `layoutZigZagContainerTopToBottom` function but use the type to determine next pos and the item positions?

dillydill123:
I think the feature is acceptable. I posted my findings

src/main/java/inventorysetups/InventorySetupsZigZagTypeID.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public enum InventorySetupsZigZagTypeID {
+    // Use if you want to withdraw using a top to bottom zigzag pattern
+    TOP_TO_BOTTOM("Top to Bottom", 0),
+
+    // Use if you want to withdraw using a bottom to top zigzag pattern
+    Bottom_To_Top("Bottom to Top", 1);
```
dillydill123, line 12: `BOTTOM_TO_TOP`

src/main/java/inventorysetups/InventorySetupLayoutUtilities.java
```diff
@@ -153,8 +156,14 @@ private int layoutZigZagContainer(final List<InventorySetupsItem> container, fin
 
 		int doubleRowStart = 0;
 		int nextPos = 0;
+		boolean topToBottom = config.zigZagType().equals(InventorySetupsZigZagTypeID.TOP_TO_BOTTOM);
 		final int rowSize = 8;
 
+		if (!topToBottom)
+		{
+			nextPos = 8;
```
dillydill123, line 164: use a ternary operator to set `nextPos` on initialization based on `topToBottom`

src/main/java/inventorysetups/InventorySetupLayoutUtilities.java
```diff
@@ -97,12 +97,15 @@ public Layout getZigZagLayout(final InventorySetup setup, final boolean addToTag
 		layout.resize(newSizeGuess);
 		final HashMap<Integer, Integer> counter = new HashMap<>();
 
+		InventorySetupsZigZagTypeID zigZagType = config.zigZagType();
```
dillydill123, line 100: It doesn't look like this is used at all here?

src/main/java/inventorysetups/InventorySetupLayoutUtilities.java
```diff
@@ -97,12 +97,15 @@ public Layout getZigZagLayout(final InventorySetup setup, final boolean addToTag
...
 
 		int nextPos = layoutZigZagContainer(setup.getEquipment(), layout, tag, addToTag, startOfEquipment, counter);
+
```
dillydill123, line 103: Can remove this extra new line created

src/main/java/inventorysetups/InventorySetupLayoutUtilities.java
```diff
@@ -97,12 +97,15 @@ public Layout getZigZagLayout(final InventorySetup setup, final boolean addToTag
...
 		if (setup.getQuiver() != null && !setup.getQuiver().isEmpty())
 		{
 			addItemToLayout(layout, tag, setup.getQuiver().get(0), nextPos, addToTag, counter);
 		}
+
```
dillydill123, line 108: Same here, can remove this extra new line

dillydill123:
Looks much better, last few things

src/main/java/inventorysetups/InventorySetupLayoutUtilities.java
```diff
@@ -163,21 +164,43 @@ private int layoutZigZagContainer(final List<InventorySetupsItem> container, fin
...
 			{
-				// We are in the bottom half of a double. Go back up and add one to move to the right.
-				nextPos = (nextPos - rowSize) + 1;
+				if (nextPos == rowSize - 1)
+				{
+					// We hit the end of a double row, we need to start a new one.
+					doubleRowStart += 2;
+					nextPos = (doubleRowStart * rowSize) - 1;
+				}
+				if (nextPos > ((doubleRowStart * rowSize) + rowSize) - 1)
```
edmeyer8851, line 194: swapped this back to if instead of else if as well, the change introduced a bug with BottomToTop layouts that I discovered while regression testing

dillydill123: I'm not sure about this. I would expect the same bug with TopToBottom then since it's using the same control flow logic.

I think there's actually a slight bug (from me) in the first `if` statement. It's only ever considering one "Double row". Not sure if this is the bug you are running into but because no container is larger than 32 items (unless someone put 32 items in additional filtered items) the bug probably hasn't been hit much.

src/main/java/inventorysetups/InventorySetupLayoutUtilities.java
```diff
@@ -163,21 +164,43 @@ private int layoutZigZagContainer(final List<InventorySetupsItem> container, fin
...
+				{
+					// We are in the bottom half of a double. Go back up and add one to move to the right.
+					nextPos = (nextPos - rowSize) + 1;
+				}
+            }
 			else
 			{
-				// We are in the bottom half of a double. Go back up and add one to move to the right.
-				nextPos = (nextPos - rowSize) + 1;
+				if (nextPos == rowSize - 1)
```
dillydill123, line 188: Should the logic here be `(nextPos == ((doubleRowStart * rowSize) + rowSize - 1))`?

src/main/java/inventorysetups/InventorySetupLayoutUtilities.java
```diff
@@ -163,21 +164,43 @@ private int layoutZigZagContainer(final List<InventorySetupsItem> container, fin
...
+				{
+					// We hit the end of a double row, we need to start a new one.
+					doubleRowStart += 2;
+					nextPos = (doubleRowStart * rowSize) - 1;
```
dillydill123, line 192: This is starting at the wrong spot. I think it should be `nextPos = (doubleRowStart * rowSize) + rowSize)`. That is it will start on the bottom of the double row. Maybe that's your bug?

dillydill123:
Tried to do math to help you out. Hopefully it makes sense. Double check my math for me

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -764,6 +765,8 @@ private void previewNewLayout(final InventorySetup setup, final InventorySetupLa
 		{
 			final Layout old = layoutUtilities.getSetupLayout(setup);
 
+			InventorySetupsZigZagTypeID zigZagType = config.zigZagType();
```
dillydill123, line 768: Looks like this is not needed?

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -145,6 +145,7 @@ public class InventorySetupsPlugin extends Plugin
 	public static final String CONFIG_KEY_PERSIST_HOTKEYS = "persistHotKeysOutsideBank";
 	public static final String CONFIG_KEY_USE_LAYOUTS = "useLayouts";
 	public static final String CONFIG_KEY_LAYOUT_DEFAULT = "defaultLayout";
+	public static final String CONFIG_KEY_ZIGZAG_TYPE = "zigZagType";
```
dillydill123, line 148: Looks like this is not needed?

PR #389, opened 2026-01-17
Fix infinite loop error deserializing legacy setups

### LINKED ISSUE #388: RuneLite freeze on startup when loading Inventory Setups V3 config containing legacy "s" field
RuneLite can hard-freeze with no error log during plugin startup/config load. The freeze happens inside gson.fromJson() while deserializing a V3 setup that contains item entries like {"id":19564,"s":0}.

Here's the inventory setup it tried to load at line
https://github.com/dillydill123/inventory-setups/blob/master/src/main/java/inventorysetups/InventorySetupsPersistentDataManager.java#L196

Not familiar with the plugin's code but I imagine it just didn't handle migrating my saved data well?

```json
{
  "inv": [
    {
      "id": 19564,
      "s": 0
    },
    {
      "id": 6714,
      "s": 1
    },
    {
      "id": 13139,
      "s": 2
    },
    {
      "id": 2434,
      "s": 3
    },
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    null,
    {
      "id": 22986,
      "s": 24
    },
    {
      "id": 560,
      "q": 44940,
      "s": 25
    },
    {
      "id": 562,
      "q": 93278,
      "s": 26
    },
    {
      "id": 555,
      "q": 120423,
      "s": 27
    }
  ],
  "eq": [
    {
      "id": 10456,
      "s": 0
    },
    {
      "id": 21795,
      "s": 1
    },
    {
      "id": 12002,
      "s": 2
    },
    {
      "id": 21006,
      "s": 3
    },
    {
      "id": 21021,
      "s": 4
    },
    {
      "id": 25985,
      "s": 5
    },
    null,
    {
      "id": 21024,
      "s": 7
    },
    null,
    {
      "id": 19544,
      "s": 9
    },
    {
      "id": 22954,
      "s": 10
    },
    null,
    {
      "id": 12601,
      "s": 12
    },
    {
      "id": 22945,
      "s": 13
    }
  ],
  "name": "Bursting Monkeys",
  "hc": "#FFFF0000",
  "hd": true,
  "fb": true,
  "uh": true,
  "sb": 1
}
```

FiveNine:
Fixes #388 .

### DIFF
--- a/src/main/java/inventorysetups/serialization/InventorySetupItemSerializableTypeAdapter.java
+++ b/src/main/java/inventorysetups/serialization/InventorySetupItemSerializableTypeAdapter.java
@@ -5,9 +5,11 @@ import com.google.gson.stream.JsonReader;
 import com.google.gson.stream.JsonToken;
 import com.google.gson.stream.JsonWriter;
 import inventorysetups.InventorySetupsStackCompareID;
+import lombok.extern.slf4j.Slf4j;
 
 import java.io.IOException;
 
+@Slf4j
 public class InventorySetupItemSerializableTypeAdapter extends TypeAdapter<InventorySetupItemSerializable>
 {
 	@Override
@@ -82,6 +84,7 @@ public class InventorySetupItemSerializableTypeAdapter extends TypeAdapter<Inven
 						break;
 					default:
 						// Handle any issues from legacy migrations without getting stuck in infinite loops
+						log.warn("Skipping unknown field '{}' in InventorySetupItemSerializable", fieldName);
 						in.skipValue();
 						break;
 				}
@@ -89,6 +92,7 @@ public class InventorySetupItemSerializableTypeAdapter extends TypeAdapter<Inven
 			else
 			{
 				// Defensive: ensure we always make progress even if we somehow end up on a value token.
+				log.warn("Skipping unknown token '{}' in InventorySetupItemSerializable", token);
 				in.skipValue();
 			}
 		}


src/main/java/inventorysetups/serialization/InventorySetupItemSerializableTypeAdapter.java
```diff
@@ -81,12 +81,20 @@ public InventorySetupItemSerializable read(JsonReader in) throws IOException
...
 					default:
+						// Handle any issues from legacy migrations without getting stuck in infinite loops
+						in.skipValue();
 						break;
 				}
 			}
+			else
+			{
+				// Defensive: ensure we always make progress even if we somehow end up on a value token.
+				in.skipValue();
```
dillydill123: Please add a logger warning to indicate we skipped something

src/main/java/inventorysetups/serialization/InventorySetupItemSerializableTypeAdapter.java
```diff
@@ -81,12 +81,20 @@ public InventorySetupItemSerializable read(JsonReader in) throws IOException
 						sc = InventorySetupsStackCompareID.valueOf(in.nextString());
 						break;
 					default:
+						// Handle any issues from legacy migrations without getting stuck in infinite loops
+						in.skipValue();
```
dillydill123: Please add a logger warning to indicate we skipped something

FiveNine:
thank you for the quick response :smile:

dillydill123:
No problem, thanks for the PR. I will do some additional testing on my side. This is a critical part of the plugin so it deserves some extra care. It might be some time to see this in a release. If you are blocked from using on your end, you can use a separate profile for now and back up the "bad" data. Or hack the plugin to work for you in the short term since you seem to know what you're doing.

FiveNine:
I managed to go into the profile's properties and cut out the `"s"` parts in each setup. Seems to be working smoothly now.

But I also found the culprit that made this happen. Had a friend's project setup on my PC a week ago when he was showing me his WIP plugin irl. I just inspected his setup and he had been fiddling with this plugin in the same project which caused the unexpected "s" field to pop into the saved setup.

Up to you whether you want to merge this PR later but I feel like this is just a good safety to add in as causing RL to freeze on a malformed config doesn't sound good at all. :eyes:

Issue #225, opened 2023-08-18
Suggestion: Allow filters to work for group storage


edlloyd8:
Would be nice so I don't have to scan the group storage for the gear I need for my Hydra trip :) (I always forget something)

PJGJ210:
watching

dillydill123:
Does anyone know if bang tags works with the group storage? I can take a look

wouterrutgers:
Hi @dillydill123 👋

Just wanted to chime in and say I'd really love to see this feature added! Do you happen to have a group ironman setup to test with?

I've got basic Java knowledge, and I can run RuneLite plugins locally, so I'd be happy to help test things out if needed!

Thanks again for all your work!

dillydill123:
Thanks, but no work has been done on this. It might be a large task that needs support from core RL, especially since Inventory Setups is using bank tags to do the filtering and layouts.

wouterrutgers:
Thanks for the update. That's unfortunate, but I understand the constraints.. If things change in the future or core RL adds support, this would still be very cool to have added and I'm happy to help out at that point! Thanks again for all your work on the plugin.

PR #411, opened 2026-06-20
Implement basic filtering to GIM storage

Robbejj:
Extends the existing bank-filtering feature to the Group Ironman shared storage. RuneLites core Bank Tags (which powers the normal-bank filter) only works on the regular bank, so this adds a small, custom "classic" filter for the shared bank: when a filtering-enabled setup is selected, items not in the setup are hidden. Matching items are left in place (no repositioning), so withdrawing always targets the correct storage slot.

So it is not as clean as the main bank filtering is, but in my opinion its a good start.

https://github.com/user-attachments/assets/7ffdca4e-5f1d-44cc-a343-0f861a39423f

Would solve #225

### DIFF
--- a/src/main/java/inventorysetups/InventorySetupsConfig.java
+++ b/src/main/java/inventorysetups/InventorySetupsConfig.java
@@ -31,6 +31,7 @@ import static inventorysetups.InventorySetupsPlugin.CONFIG_KEY_PERSIST_HOTKEYS_C
 import static inventorysetups.InventorySetupsPlugin.CONFIG_KEY_USE_OLD_ITEM_SEARCH;
 import static inventorysetups.InventorySetupsPlugin.CONFIG_KEY_ZIGZAG_TYPE;
 import static inventorysetups.InventorySetupsPlugin.CONFIG_KEY_LAYOUT_DUPLICATES;
+import static inventorysetups.InventorySetupsPlugin.CONFIG_KEY_FILTER_SHARED_BANK;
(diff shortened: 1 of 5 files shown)


Robbejj:
@dillydill123 Any thoughts on this? Would love to see it live as its real useful for any GIM using the plugin.

dillydill123:
Hey! Thanks for your effort and time on this. I'm open to this and will take a look.

Robbejj:
Let me know if any changes are wanted and I'll fix 👍
I've been testing this for few days while doing slayer etc and I haven't noticed any issues with it myself.

dillydill123:
@Robbejj, Have you considered adding this to bank tags? Most of the filtering logic in inventory setups is handled by bank tags. A solution in that plugin would be better suited and likely require no or little integration in Inventory Setups to work.

Robbejj:
@dillydill123 I thought about that when I wrote this but to be honest I'm not a fan of trying to make changes into any **default** plugin since it would probably take a year before anyone even takes a look at the PR..
Even if the bank tags worked parts of this PR would still be needed so this plugin loads when the shared bank is open, would only need to edit the filter.
I can look into opening that PR.

Robbejj:
https://github.com/runelite/runelite/pull/20270

dillydill123:
@Robbejj It looks like https://github.com/runelite/runelite/pull/20270 was merged. What's missing from Inventory Setups side?

dillydill123:
Closing now that the hub plugin and plugin message api are available

Issue #415, opened 2026-06-28
[Feature Request] : PluginMessage API for inter-plugin integration


mepatrick73:
Hey, I'm working on a radial ring menu plugin for the hub that lets you bind actions to a hotkey-triggered radial menu. One of the integrations I want to add is Inventory Setups so you can switch setups directly from the ring.

Since hub plugins can't use reflection I was looking at net.runelite.client.events.PluginMessage as the way to do this. I'd need two things from inventory setups:

1. Broadcasting the current setup list on startup and whenever setups are added, removed, or renamed under namespace inventory-setups, name setups-changed, with the list of setup names in the data map.
2. Handling a view message under the same namespace that takes a setup name and switches the panel to that setup.

On my end I'd send something like:
```java
eventBus.post(new PluginMessage("inventory-setups", "view", Map.of("setup", setupName)));
```

Let me know if this is something you'd be open to and I'm happy to help however I can.

dillydill123:
This sounds interesting. Do you have some mock up of how it would look like? Some people have hundreds of setups.

My only concern is what performance impact does this have. It doesn't sound too bad, since it's on less frequent events (adding/removing/renaming), but is something to consider.

mepatrick73:
I do have some mock up of the plugin I'm making if you wanna take look: 
https://github.com/mepatrick73/ring-menu-plugin/tree/master

How should I go about making an integration? Should I fork this repo and create a pull request with my feature request?

dillydill123:
Yeah, create a fork. How are you currently getting this info? You could also potentially subscribe to config changes and hook into when `inventorysetups` but I could see this PluginMessage being a more "official" way for other plugins to interact with inv setups. Are there any other plugins that do this?

mepatrick73:
I've opened a PR for this: #416

To your questions:

- **How I get the data now:** reading the config directly (enumerating the `setupsV3_` keys, parsing the name, then opening the layout bank tag via the murmur3 hash). It works, but it couples me to your v3 schema and the tag hashing, which is exactly what the comment in `updateConfig` was written to avoid.
- **Config subscription vs PluginMessage:** subscribing to `ConfigChanged` keeps the same coupling, since I'd still need your key layout and hashing. PluginMessage lets you own the contract.
[...]

PR #416, opened 2026-07-01
Add PluginMessage API for inter-plugin integration

mepatrick73:
# Add PluginMessage API for inter-plugin integration

Addresses #415.

## Summary

Adds a `PluginMessage` API under the `inventory-setups` namespace so other plugins can list setups and open or clear them, without reading the config keys or reimplementing the layout tag hashing.

## Messages (namespace `inventory-setups`)

| Direction | name | data | behavior |
|---|---|---|---|
| out | `setups-changed` | `setups`: `List<String>`, `version`: `int` | broadcast on startup and whenever the set of setups changes |
| in | `get-setups` | `setups`: mutable `Collection<String>` | filled synchronously with the current setup names |
| in | `view` | `setup`: `String` | opens the setup and filters the bank, same as the worn items menu |
| in | `clear` | `setup`: `String` (optional) | clears the current setup; if a name is given, only when it is the active one |

## Performance

- The broadcast piggybacks on `updateConfig`, which already serializes every setup to config on each change, so building a `List<String>` of names next to that is negligible.
- It only fires on add, remove, rename, import and reorder, never per tick or per frame, and it skips the post when the name list is unchanged, so slot and note edits do not broadcast.
- The build and post run on the client thread, and `get-setups` is answered from a `volatile` immutable snapshot, so it is safe to call from any thread.

## Versioning

`setups-changed` carries a `version` field (`API_VERSION`, currently 1) so consumers can detect future breaking changes.

## Testing

Verified end to end against a consumer plugin (a radial ring menu): startup broadcast, live add / rename / delete updates, bank filtering, and name-scoped clear, including both load orders (consumer first and inventory setups first).

### DIFF
--- a/src/main/java/inventorysetups/InventorySetupsPlugin.java
+++ b/src/main/java/inventorysetups/InventorySetupsPlugin.java
@@ -181,24 +181,6 @@ public class InventorySetupsPlugin extends Plugin
 	private static final int SPELLBOOK_VARBIT = 4070;
 	private static final int BANK_TAG_OPTIONS = BankTagsService.OPTION_ALLOW_MODIFICATIONS | BankTagsService.OPTION_HIDE_TAG_NAME;
 
-	// PluginMessage API for other plugins to list and open/clear inventory setups. See #415.
-	public static final String API_NAMESPACE = "inventory-setups";
-	// Bumped when the contract below changes in a breaking way. Shipped in setups-changed as data["version"].
-	public static final int API_VERSION = 1;
-	// out: broadcast when the setups change. data["setups"] = List<String> of names, data["version"] = int.
-	public static final String API_MSG_SETUPS_CHANGED = "setups-changed";
-	// in: list setups on demand (for plugins that start after us). Put a mutable Collection<String> under
-	// "setups"; it is filled synchronously with the current setup names.
-	public static final String API_MSG_GET_SETUPS = "get-setups";
-	// in: open a setup, filtering the bank like the worn items menu. data["setup"] = name.
-	public static final String API_MSG_VIEW = "view";
-	// in: clear the current setup (like worn items "Close current setup"). data["setup"] = name to clear
-	// only when it is the active setup; omit to clear whatever is active.
-	public static final String API_MSG_CLEAR = "clear";
-	public static final String API_DATA_SETUPS = "setups";
-	public static final String API_DATA_SETUP = "setup";
-	public static final String API_DATA_VERSION = "version";
-
 	@Inject
 	@Getter
 	private Client client;
@@ -306,6 +288,9 @@ public class InventorySetupsPlugin extends Plugin
 	@Getter
 	private InventorySetupsAmmoHandler ammoHandler;
 
+	@Getter
+	private InventorySetupsPluginMessageHandler pluginMessageHandler;
+
 	// Used to defer highlighting to GameTick
 	private boolean shouldTriggerInventoryHighlightOnGameTick;
 
@@ -380,6 +365,7 @@ public class InventorySetupsPlugin extends Plugin
 		this.sections = new ArrayList<>();
 		this.dataManager = new InventorySetupsPersistentDataManager(this, configManager, cache, gson, inventorySetups, sections);
 		this.ammoHandler = new InventorySetupsAmmoHandler(this, client, itemManager, panel, config);
+		this.pluginMessageHandler = new InventorySetupsPluginMessageHandler(this, clientThread, eventBus, panel);
 		this.layoutUtilities = new InventorySetupLayoutUtilities(itemManager, tagManager, layoutManager, config, client);
 		this.canUseLayouts = canUseLayouts();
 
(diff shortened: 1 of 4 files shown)


src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -2654,4 +2678,107 @@ public void updateSectionName(final InventorySetupsSection section, final String
...
+				clientThread.invoke(() ->
+				{
+					final InventorySetup current = panel.getCurrentSelectedSetup();
+					if (current == null)
+					{
+						return;
+					}
+					// If a setup name is given, only clear when it is the one currently shown, so a caller
+					// never closes a setup the user switched to themselves.
+					if (nameObj instanceof String && !current.getName().equals(nameObj))
```
dillydill123: Is the idea that `nameObj` will be `null` if the user just wants to clear the current setup? I think two if statements are better here (one to check for `nameObj == null` perhaps) for readability.

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -2654,4 +2678,107 @@ public void updateSectionName(final InventorySetupsSection section, final String
...
+		if (!API_NAMESPACE.equals(message.getNamespace())
+			|| API_MSG_SETUPS_CHANGED.equals(message.getName()))
+		{
+			// Ignore other namespaces and our own outgoing broadcast.
+			return;
+		}
+
+		switch (message.getName())
+		{
+			case API_MSG_GET_SETUPS:
```
dillydill123: Add a default case and log that a message is not supported was ignored.

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -2654,4 +2678,107 @@ public void updateSectionName(final InventorySetupsSection section, final String
...
+			setupNamesSnapshot = names;
+			eventBus.post(new PluginMessage(API_NAMESPACE, API_MSG_SETUPS_CHANGED,
+				Map.of(API_DATA_SETUPS, names, API_DATA_VERSION, API_VERSION)));
+		});
+	}
+
+	@Subscribe
+	public void onPluginMessage(final PluginMessage message)
+	{
+		if (!API_NAMESPACE.equals(message.getNamespace())
```
dillydill123: This is probably easier read if split into two if statements

dillydill123 (commented):
Is there any information about how PluginMessage works? I think I'm a little out of my breath in this review and need to do some digging. In the meantime I've made some comments. 

One additional point of feedback: Can you move this logic to a new `InventorySetupsPluginMessageHandler.java` or similar? `InventorySetupsPlugin` is already a god object doing too much. The plugin can subscribe to the message but then call the plugin message handler to do all the logic. This would keep the API constants and volatile member in their own namespace.

dillydill123:
>It only fires on add, remove, rename, import and reorder, never per tick or per frame, and it skips the post when the name list is unchanged, so slot and note edits do not broadcast.

~~Based on where it is being fired in `updateConfig`, the above statement is incorrect. It will fire when slot and note edits are made from what I understand..~~ NVM, looks fine

mepatrick73:
Thanks for the review, all addressed in cde3f31:

- Moved all the API logic into a new `InventorySetupsPluginMessageHandler`, following the same pattern as `InventorySetupsAmmoHandler` (constructed in `startUp`). The plugin now only has a one-line `@Subscribe` that delegates, plus a one-line `broadcastSetupsChanged()` forward for the data manager, so the API constants and the volatile snapshot live in their own class.
- Split the namespace / own-broadcast guard into two `if` statements.
- Added a `default` case that logs unsupported message names at debug.
- Reworked the `clear` name check into explicit branches. And yes, exactly as you guessed: a `null` name means "clear whatever is active", while a provided name only clears when it matches the currently shown setup, so a caller never closes a setup the user switched to themselves.

**On how `PluginMessage` works:** it's a core RuneLite event ([`net.runelite.client.events.PluginMessage`](https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/events/PluginMessage.java)) added by Adam in [runelite/runelite@0b12333](https://github.com/runelite/runelite/commit/0b12333eafcb9ae1b394849fd65b0526a4e3a8f9) specifically so hub plugins can talk to each other despite living in separate classloaders. Per its javadoc, it is "an event pluginhub plugins can use to send data to each other." It's just an immutable `namespace` + `name` + `Map<String, Object>` posted on the normal `EventBus`: no serialization or IPC, and `eventBus.post` invokes subscribers synchronously on the calling thread (each wrapped in its own try/catch, so a bad subscriber can't hurt the poster). That synchronous delivery is what makes `get-setups` work, since the caller's collection is filled before its own `post()` returns. The one constraint is that payload objects must be types both classloaders can see; this API only uses `String`, `int` and `Collection<String>`. The [Shortest Path plugin documents the same pattern](https://github.com/Skretzo/shortest-path/wiki/Cross-plugin-communication) (`"shortestpath"` namespace with `path`/`clear` messages), which is what this follows.

Retested the whole surface end to end after the refactor (both load orders, broadcasts on add/rename/reorder/delete, no broadcast on slot and note edits, view with bank filtering, and both the matching and non-matching clear cases). Behavior is identical to what you reviewed.

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -0,0 +1,178 @@
...
+				break;
+			}
+			case API_MSG_CLEAR:
+			{
+				handleClear(message);
+				break;
+			}
+			default:
+			{
+				log.debug("Ignoring unsupported message '{}' in the {} namespace", message.getName(), API_NAMESPACE);
```
dillydill123: nit: I'd make this a warning

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -0,0 +1,178 @@
...
+import net.runelite.client.callback.ClientThread;
+import net.runelite.client.eventbus.EventBus;
+import net.runelite.client.events.PluginMessage;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+// PluginMessage API for other plugins to list and open/clear inventory setups. See #415.
```
dillydill123, line 14: Perhaps add a blurb here how to use this if you're another plugin? Also would be good to link to this in the README under a new subsection in `Other Plugin Compatibilities`

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -0,0 +1,178 @@
...
+				log.debug("Ignoring unsupported message '{}' in the {} namespace", message.getName(), API_NAMESPACE);
+				break;
+			}
+		}
+	}
+
+	private void handleGetSetups(final PluginMessage message)
+	{
+		final Object container = message.getData().get(API_DATA_SETUPS);
+		if (container instanceof Collection)
```
dillydill123, line 123: Should we add some `else` logic with a warning if the user does not pass a `Collection`?

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -0,0 +1,178 @@
...
+			((Collection<String>) container).addAll(setupNamesSnapshot);
+		}
+	}
+
+	private void handleView(final PluginMessage message)
+	{
+		final Object nameObj = message.getData().get(API_DATA_SETUP);
+		if (!(nameObj instanceof String))
+		{
+			return;
```
dillydill123, line 142: nit: log a warning here

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -0,0 +1,178 @@
...
+		// Resolve and apply on the client thread, where the setups are otherwise accessed.
+		clientThread.invoke(() ->
+		{
+			final InventorySetup target = plugin.getInventorySetups().stream()
+				.filter(setup -> setup.getName().equals(targetName))
+				.findFirst()
+				.orElse(null);
+			if (target == null)
+			{
+				log.debug("Ignoring view request for unknown setup '{}'", targetName);
```
dillydill123: I'd make this a warning as well

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -0,0 +1,178 @@
...
+			{
+				return;
+			}
+			if (nameObj == null)
+			{
+				// No name given: clear whatever setup is active.
+				panel.returnToOverviewPanel(false);
+				return;
+			}
+			// A name was given: only clear when it is the setup currently shown, so a caller never
```
dillydill123, line 182: Should we do a an `instanceof` check with `String` here as well?

dillydill123 (commented):
Looks much better, I'll take a look at shortest path. Some minor nits in the meantime

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -0,0 +1,178 @@
...
+			{
+				log.debug("Ignoring unsupported message '{}' in the {} namespace", message.getName(), API_NAMESPACE);
+				break;
+			}
+		}
+	}
+
+	private void handleGetSetups(final PluginMessage message)
+	{
+		final Object container = message.getData().get(API_DATA_SETUPS);
```
dillydill123: Based on shortest path plugin it seems we should use `getOrDefault(..., null)` to protect against the message data not having the key.

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -0,0 +1,178 @@
...
+			// eventBus.post is synchronous, so the caller's collection is filled before its own
+			// post() call returns.
+			//noinspection unchecked
+			((Collection<String>) container).addAll(setupNamesSnapshot);
+		}
+	}
+
+	private void handleView(final PluginMessage message)
+	{
+		final Object nameObj = message.getData().get(API_DATA_SETUP);
```
dillydill123: `getOrDefault(..., null)` as discussed above

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -0,0 +1,178 @@
...
+				log.debug("Ignoring view request for unknown setup '{}'", targetName);
+				return;
+			}
+			panel.setCurrentInventorySetup(target, true);
+		});
+	}
+
+	private void handleClear(final PluginMessage message)
+	{
+		final Object nameObj = message.getData().get(API_DATA_SETUP);
```
dillydill123: `getOrDefault(..., null)` as discussed above

mepatrick73:
Thanks for the second pass, all addressed in eeadeaa:

- Bumped all four of the ignore paths to `log.warn` (unsupported message name, non-Collection `get-setups` payload, non-String `view` name, unknown setup name).
- Added the `instanceof String` check to `clear`; a non-String name is now rejected with a warning, while `null` keeps its "clear whatever is active" meaning.
- Switched the payload reads to `getOrDefault(..., null)` per Shortest Path. Worth noting `Map.get` already returns `null` for a missing key, so this is a style match rather than a behavior change.
- Expanded the class comment into a short how-to with a code example, and added a "PluginMessage API for Plugin Developers" subsection to the README under Other Plugin Compatibilities.

dillydill123:
Nice, looks like we are almost there. I think there is one case you need to test. What happens if a user changes profiles? There is a function `switchProfile` that has a handler for this. I think you need to broadcast setups there as the config will totally have changed and subscribers of the plugin message will need to be informed.

mepatrick73:
Good catch, fixed in ba24084. `switchProfile` now calls `broadcastSetupsChanged()` right after `dataManager.loadConfig()`, mirroring what `startUp` already does, so subscribers are informed when the config changes out from under them.

Tested by switching between three profiles with different setup lists (269, 65, and 0 setups). Each switch fired one `setups-changed` broadcast with the new profile's names, verified with temporary debug logging on the broadcast path. Switching between two profiles whose setup names are identical correctly skipped the post thanks to the existing unchanged-names guard, so integrating plugins don't get a redundant event.

dillydill123 (approved):
Thanks for your work and patience (been a busy summer). Will look to get this in the next update within two weeks. There are some other things I want to include in the next update.

PR #418, opened 2026-07-06
Fix translucent highlight ghosting and add live color preview

Darkforge317:
> [!IMPORTANT]
> @jarromie had originally re-implemented support for Alpha/Opacity in highlights.
> This PR combines their work with a fix for a merge-blocking bug + a few Quality of Life updates for color selection.

# Summary
- Fix visual ghosting and alpha stacking when using semi-transparent highlight colors on inventory setup slots
- Re-enable alpha support in the highlight and display color pickers (previously hard-coded to 255 due to the bug we fix here)
- Refresh slot highlighting live while the color picker is open and the same setup is being viewed
- Add a "Change highlight color" button above setups that have highlighting enabled

# The Ghosting Problem

Translucent highlight colors were applied via `setBackground()` on opaque `JPanel` slots (`InventorySetupsSlot`). Swing does not handle this correctly: semi-transparent fills blend over stale buffer pixels instead of clearing them first. 

This caused:

- Ghosted item icons and RuneLite UI bleeding into slots (especially on first open after changing alpha)
- Highlight opacity accumulating on repeated `doHighlighting()` calls (e.g. alpha 150 layered twice appearing nearly solid)
- Artifacts that cleared after scrolling or re-opening a setup

Alpha had been disabled in the color picker callbacks as a workaround.
617264197-d6eb798e-0814-4fae-a088-04ee1903ed24

## Solution

**Ghosting fix (`InventorySetupsSlot`):**
- `setOpaque(false)` so Swing does not auto-fill with a broken translucent background
- Custom `paintComponent()` that always fills the slot's opaque base color first, then draws the highlight color in a single paint pass
- `doResetHighlight()` restores each slot's constructor-provided default background
- 
**Transparency re-enabled:**
- Highlight color picker passes the full picked `Color` (including alpha)
- Display color picker preserves alpha via `c.getAlpha()`

# Quality of Life Updates
## Live color preview
- When the highlight color changes and that setup is already open in the detail view, re-run highlighting immediately so slots update without closing/reopening the setup
- The paint bucket underline has been given the same live updating behavior

## In-Setup Color Picker
- Added a "Change highlight color" element that opens the color picker. This provides a better user experience than needing to back out of a setup to change its coloring. 
- This element only shows if highlighting is enabled for the current setup, otherwise it is hidden.
- Color picker now opens to the left of the cursor. Creation now handled by `positionColorPickerLeftOfCursor`. This was done to prevent the color picker from being created half off-screen and half on-screen. 

# Live Demo

https://github.com/user-attachments/assets/5265ec4c-0f6a-4a63-92db-6ce7434790dd

### DIFF
--- a/src/main/java/inventorysetups/ui/InventorySetupsSlot.java
+++ b/src/main/java/inventorysetups/ui/InventorySetupsSlot.java
@@ -34,13 +34,14 @@ import inventorysetups.InventorySetupsVariationMapping;
 import lombok.Getter;
 import lombok.Setter;
 import net.runelite.client.game.ItemManager;
-import net.runelite.client.ui.ColorScheme;
 import net.runelite.client.ui.FontManager;
 import net.runelite.client.util.AsyncBufferedImage;
 
 import javax.swing.*;
 import java.awt.Color;
 import java.awt.Dimension;
+import java.awt.Graphics;
+import java.awt.Graphics2D;
 import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
 import java.awt.Insets;
@@ -75,6 +76,8 @@ public class InventorySetupsSlot extends JPanel
 	@Getter
 	private JPopupMenu shiftRightClickMenu;
 
+	private final Color defaultBackgroundColor;
+
 	public static final int SLOT_WIDTH = 46;
 	public static final int SLOT_HEIGHT = 42;
 
@@ -85,6 +88,7 @@ public class InventorySetupsSlot extends JPanel
 
 	public InventorySetupsSlot(Color color, InventorySetupsSlotID id, int indexInSlot, int width, int height)
 	{
+		this.defaultBackgroundColor = color;
 		this.slotID = id;
 		this.imageLabel = new JLabel();
 		this.parentSetup = null;
@@ -129,6 +133,11 @@ public class InventorySetupsSlot extends JPanel
 		this.imageLabel.addMouseListener(menuAdapter);
 
 		setPreferredSize(new Dimension(width, height));
+
+		// Swing assumes opaque components paint an opaque background. If the
+		// background color has an alpha < 255, previous pixels may show through.
+		// We set Slots to be non-opaque and handle painting ourselves in `PaintComponent`.
+		setOpaque(false);
 		setBackground(color);
 		setLayout(new GridBagLayout());
 
@@ -143,6 +152,34 @@ public class InventorySetupsSlot extends JPanel
 		add(imageLabel);
 		add(fuzzyIndicator, fuzzyConstraints);
 		add(stackIndicator, stackConstraints);
+	}
+
+	@Override
+	protected void paintComponent(Graphics g)
+	{
+		// Since this panel is non-opaque, Swing does not clear its background.
+		// Paint an opaque default background first so translucent background
+		// colors don't blend with pixels from previous repaints.
+		Graphics2D g2d = (Graphics2D) g.create();
+		try
+		{
+			g2d.setColor(defaultBackgroundColor);
+			g2d.fillRect(0, 0, getWidth(), getHeight());
+
+			Color bg = getBackground();
+			if (bg != null && !bg.equals(defaultBackgroundColor))
+			{
+				g2d.setColor(bg);
+				g2d.fillRect(0, 0, getWidth(), getHeight());
+			}
+		}
+		finally
+		{
+			g2d.dispose();
+		}
+
+		// Call super last to paint images and text over the now correctly filled-in background
+		super.paintComponent(g);
 	}
 
 	public void setImageLabel(String toolTip, BufferedImage itemImage, boolean isFuzzy, InventorySetupsStackCompareID stackCompare)
@@ -399,7 +436,7 @@ public class InventorySetupsSlot extends JPanel
 
 	public static void doResetHighlight(final InventorySetupsSlot containerSlot)
 	{
-		containerSlot.setBackground(ColorScheme.DARKER_GRAY_COLOR);
+		containerSlot.setBackground(containerSlot.defaultBackgroundColor);
 	}
 
 	public static boolean shouldHighlightSlotBasedOnStack(final InventorySetupsStackCompareID stackCompareType, final Integer savedItemQty, final Integer currItemQty)
--- a/src/main/java/inventorysetups/ui/InventorySetupsNameActions.java
+++ b/src/main/java/inventorysetups/ui/InventorySetupsNameActions.java
@@ -328,7 +328,7 @@ public class InventorySetupsNameActions<T extends InventorySetupsDisplayAttribut
 					plugin.openColorPicker("Choose a Display color", currentDisplayColor == null ? JagexColors.MENU_TARGET : currentDisplayColor,
 							c ->
 							{
-								Color newColor = new Color(c.getRed(), c.getGreen(), c.getBlue(), 255);
+								Color newColor = new Color(c.getRed(), c.getGreen(), c.getBlue(), c.getAlpha());
 								updateDisplayColorLabel(newColor);
 							}
 					);
--- a/src/main/java/inventorysetups/ui/InventorySetupsStandardPanel.java
+++ b/src/main/java/inventorysetups/ui/InventorySetupsStandardPanel.java
@@ -317,9 +317,7 @@ public class InventorySetupsStandardPanel extends InventorySetupsPanel implement
 					plugin.openColorPicker("Choose a Highlight color", invSetup.getHighlightColor(),
 						c ->
 						{
-							// Don't consider the transparency component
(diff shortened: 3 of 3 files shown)


jarromie:
Awesome! Glad you were able to make time for this, as it would've been a while before I'd be able to investigate. Will be very happy when this makes it into the plugin hub version.

dillydill123:
Thanks for investigating this. A few questions/things to consider:
- Should this be part of RuneLite itself? All `JPanel`'s could do this logic if they are opaque. Curious what RL devs would think since it's a bit of a foundational change.
- Does this have any performance impact? I'm not actually sure how to measure but modifying such a hot code path like `paintComponent` deserves some investigation I feel.
- I'm not really a fan of the live color picker widget. The setup view screen is prime real estate. It doesn't fit to add something users may not interact with much (after setting an initial color) to that screen. The color picker has a a preview already. If we wanted to add a better preview, like an example slot with item inside, it could be put directly into the picker itself. I would argue this should be a separate PR anyways.

Darkforge317:
Thanks for the thorough review @dillydill123!

## RuneLite core
Good point, I hadn't considered pushing this up to `JPanel` itself. I don't have a strong opinion either way. Happy to keep it scoped to this plugin for now and let it get raised on the RL side separately if a maintainer thinks it's worth generalizing. The fix itself is a normal, documented Swing pattern (non-opaque + custom `paintComponent`), so it shouldn't need RL core changes to be correct here.

## Performance
**Worth noting**: before this fix, isOpaque() was true, so Swing was already running an automatic background fillRect() on every repaint via ComponentUI.update(). That cost already existed. Setting the slot non-opaque (needed to fix the ghosting) skips that automatic fill, and my override does the fill explicitly instead. For a non-highlighted slot it's a wash, one fillRect() either way. The only genuinely new cost is a highlighted slot's second fillRect() for the overlay color, which is the **<ins>minimum needed</ins>** to draw the highlight without ghosting. 

To confirm there's no hidden overhead, I ran the client through VisualVM's sampler (screenshot attached): Self Time and CPU Time both read 0.0 ms, the 370ms wall-clock is just Java2D's fillRect() doing its normal work, not extra work from the override.

image

## Live color picker widget
Fair, and agreed! I have trouble staying in-scope sometimes lol. I've pulled it out in the most recent commit, along with the live-refresh-on-picker-drag behavior it depended on, since that was really part of the same live-preview feature rather than the ghosting fix itself.

jarromie:
I quite like the live color picker, but do agree that it is maybe ideal for a separate PR (after this one is merged). It may be unnecessarily taking up prime real estate, but could be modified to take up less space or have the button put elsewhere (I tried to hypothesize a better solution, maybe a color picker icon beside the button to refresh setup?)

Great work regardless, really happy to see that this is going to make it into the plugin.

Darkforge317:
> I quite like the live color picker, but do agree that it is maybe ideal for a separate PR (after this one is merged). It may be unnecessarily taking up prime real estate, but could be modified to take up less space or have the button put elsewhere (I tried to hypothesize a better solution, maybe a color picker icon beside the button to refresh setup?)
> 
> Great work regardless, really happy to see that this is going to make it into the plugin.

Yeah I liked it too! I mainly did it because I found it annoying to have to go back and forth, back and forth, back... just to see what the colour <ins>actually</ins> looks like in the UI. Transparency changes the colour quite a bit because of the dark background.

To avoid UI clash, I made the existing colour picker open to the left of the cursor to prevent overlap. But theoretically could just make it open far left of the UI to prevent any contact with the inventory setup at all.

Regardless, sounds like the colour previewing isn't something that'll come into the plugin unless there's more demand for it.

Darkforge317:
> Thanks for investigating this. A few questions/things to consider:
> 
> * Should this be part of RuneLite itself? All `JPanel`'s could do this logic if they are opaque. Curious what RL devs would think since it's a bit of a foundational change.
> * Does this have any performance impact? I'm not actually sure how to measure but modifying such a hot code path like `paintComponent` deserves some investigation I feel.
> * I'm not really a fan of the live color picker widget. The setup view screen is prime real estate. It doesn't fit to add something users may not interact with much (after setting an initial color) to that screen. The color picker has a a preview already. If we wanted to add a better preview, like an example slot with item inside, it could be put directly into the picker itself. I would argue this should be a separate PR anyways.

Just checking back in @dillydill123, are we good to go now or did you have more concerns that needed to be addressed? 😃

src/main/java/inventorysetups/ui/InventorySetupsSlot.java
```diff
@@ -145,6 +150,30 @@ private void showPopupMenu(MouseEvent e)
 		add(stackIndicator, stackConstraints);
 	}
 
+	@Override
+	protected void paintComponent(Graphics g)
+	{
+		Graphics2D g2d = (Graphics2D) g.create();
+		try
+		{
+			g2d.setColor(defaultBackgroundColor);
```
dillydill123, line 166: We definitely need a comment here. If I understand correctly, something like the following
```
// Since this panel is non-opaque, Swing does not clear its background.
// Paint an opaque default background first so translucent background
// colors don't blend with pixels from previous repaints.
```

src/main/java/inventorysetups/ui/InventorySetupsSlot.java
```diff
@@ -145,6 +150,30 @@ private void showPopupMenu(MouseEvent e)
...
+				g2d.setColor(bg);
+				g2d.fillRect(0, 0, getWidth(), getHeight());
+			}
+		}
+		finally
+		{
+			g2d.dispose();
+		}
+
+		super.paintComponent(g);
```
dillydill123, line 182: Comment here since this is important at the bottom:
```
// Call super last to paint images and text over the now correctly filled-in background
```

src/main/java/inventorysetups/InventorySetupsPlugin.java
```diff
@@ -2609,6 +2615,46 @@ public void windowClosing(WindowEvent e)
 		colorPicker.setVisible(true);
 	}
 
+	private void positionColorPickerLeftOfCursor(final RuneliteColorPicker colorPicker)
```
dillydill123: I know this is a minor feature, but I'd prefer to keep the PR targeted to just the alpha fix. Please remove this piece

src/main/java/inventorysetups/ui/InventorySetupsSlot.java
```diff
@@ -129,6 +133,7 @@ private void showPopupMenu(MouseEvent e)
 		this.imageLabel.addMouseListener(menuAdapter);
 
 		setPreferredSize(new Dimension(width, height));
+		setOpaque(false);
```
dillydill123, line 140: We should put a comment here indicating the reason for this:
```
// Swing assumes opaque components paint an opaque background. If the
// background color has an alpha < 255, previous pixels may show through.
// We set Slots to be non-opaque and handle painting ourselves in `PaintComponent`.
```

dillydill123 (commented):
I've left some comments and request that you fixup the commits into one

src/main/java/inventorysetups/ui/InventorySetupsSlot.java
Darkforge317: Image

src/main/java/inventorysetups/ui/InventorySetupsSlot.java
Darkforge317: Image

src/main/java/inventorysetups/ui/InventorySetupsSlot.java
Darkforge317: Image

src/main/java/inventorysetups/InventorySetupsPlugin.java
Darkforge317: Done, and no worries. Thanks for keeping the PR scope in-line! I'll do better with this going forward, promise!

Darkforge317:
> I've left some comments and request that you fixup the commits into one

@dillydill123 done, and squashed to one commit as requested. I admire your commitment to keeping the repo clean and approachable for new contributors!

image


**<ins>EDIT:</ins>** 
Had to make another push as I noticed one of the files in your repo uses CRLF, and my commit had swapped it to LF which caused a file-wide diff for `\ui\InventorySetupsNameActions.java`. While the rest of the repo is in LF, and it makes sense to convert this to LF, that would be out of scope for this PR.

Darkforge317:
@dillydill123 happy Friday and weekend! I just wanted to check back in on this if you have time to do so.

dillydill123 (approved):
Thanks for the work on this and your patience (been a busy summer). I'll get an update out with this within the next two weeks. There are some other things I want to include in the next update.

Darkforge317:
> Thanks for the work on this and your patience (been a busy summer). I'll get an update out with this within the next two weeks. There are some other things I want to include in the next update.

No problem! Happy to help. 😁 Looking forward to the update.

PR #425, opened 2026-08-08
Extend the PluginMessage API

Robbejj:
### What
Extends the existing `PluginMessage` integration API (`InventorySetupsPluginMessageHandler`) with two small, purely additive messages:

- **`active-setup-changed`** (out, broadcast) — fires whenever the active setup changes, including edits to the contents of the setup that's still active (not just selection changes). Payload: `data["setup"]` = the active setup's name, or the key is absent when no setup is active.
- **`get-active-setup-contents`** (in, synchronous query) — Fills:
  - `data["hasActiveSetup"]` (`Boolean`) — false when no setup is selected or bank filtering is disabled for it.
  - `data["equipmentItemIds"]` / `data["inventoryItemIds"]` (`List<Integer>`) — the setup's equipment/inventory items by slot, `-1` for empty slots.
  - `data["additionalItemIds"]` (`Collection<Integer>`) — the setup's additional filtered items.

### Why
These additions would allow my upcoming plugin to mirror Inventory Setups setup's contents into the Group Ironman shared bank (filtering + a best-effort layout). The existing API exposes the list of setup *names* but nothing about which one is active or what's in it, so there was no way to do this without polling internals. These two messages close that gap the same way `get-setups`/`setups-changed` already do for the setup list.

### Changes
- `InventorySetupsPluginMessageHandler.java` — new constants/handlers for both messages, doc comment updated.
- `InventorySetupsPlugin.java` — `broadcastActiveSetupChanged()` delegate, next to the existing `broadcastSetupsChanged()`.
- `InventorySetupsPluginPanel.java` — calls the new broadcast from `setCurrentInventorySetup()`/`returnToOverviewPanel()`.
- `InventorySetupsPersistentDataManager.java` — also calls it from `updateConfig()`, so edits to the still-active setup broadcast too, not just selection changes.

No existing behavior changes; both messages are additive.

### DIFF
--- a/src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
+++ b/src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
@@ -22,8 +22,11 @@ import java.util.Map;
 //
 // Supported messages are documented on the constants below. To track the available setups, subscribe
 // to PluginMessage and listen for "setups-changed" broadcasts, and post "get-setups" once on startup
-// (posting is synchronous, so the collection you pass is filled before post() returns). Payload values
-// are plain JDK types (String, int, Collection<String>) so they are visible across plugin classloaders.
+// (posting is synchronous, so the collection you pass is filled before post() returns). To track the
+// active setup (including live edits to it), listen for "active-setup-changed" broadcasts and post
+// "get-active-setup-contents" whenever you need its current contents. Payload values are plain JDK
+// types (String, int, boolean, Collection<Integer>/<String>) so they are visible across plugin
+// classloaders.
 @Slf4j
 public class InventorySetupsPluginMessageHandler
 {
@@ -40,9 +43,22 @@ public class InventorySetupsPluginMessageHandler
 	// in: clear the current setup (like worn items "Close current setup"). data["setup"] = name to clear
 	// only when it is the active setup; omit to clear whatever is active.
 	public static final String API_MSG_CLEAR = "clear";
+	// out: broadcast when the active setup changes - on opening, closing and editing
+	// of the setup that's still active (add/remove an item, toggle fuzzy, etc.)
+	// data["setup"] = the active setup's name; the key is absent when setup is closed.
+	public static final String API_MSG_ACTIVE_SETUP_CHANGED = "active-setup-changed";
+	// in: get the active setup's contents by slot, e.g. for a plugin that wants to mirror its layout elsewhere.
+	// Put mutable Collection<Integer> under "equipmentItemIds" (EquipmentInventorySlot order,
+	// size 14), "inventoryItemIds" (size 28), and "additionalItemIds" (no position semantics). Posting is synchronous.
+	// data["hasActiveSetup"] is set to false when no setup is active or the active setup has bank filtering disabled.
+	public static final String API_MSG_GET_ACTIVE_SETUP_CONTENTS = "get-active-setup-contents";
 	public static final String API_DATA_SETUPS = "setups";
 	public static final String API_DATA_SETUP = "setup";
 	public static final String API_DATA_VERSION = "version";
+	public static final String API_DATA_HAS_ACTIVE_SETUP = "hasActiveSetup";
+	public static final String API_DATA_EQUIPMENT_ITEM_IDS = "equipmentItemIds";
+	public static final String API_DATA_INVENTORY_ITEM_IDS = "inventoryItemIds";
+	public static final String API_DATA_ADDITIONAL_ITEM_IDS = "additionalItemIds";
 
 	private final InventorySetupsPlugin plugin;
 	private final ClientThread clientThread;
@@ -80,13 +96,26 @@ public class InventorySetupsPluginMessageHandler
 		});
 	}
 
+	// Broadcasts the active setup's name, or clears it. Called on selection change and on every content
+	// edit of the setup that's still active. Deliberately not deduped like broadcastSetupsChanged() -
+	// content edits don't change the name, but callers still need to know to re-read the contents.
+	public void broadcastActiveSetupChanged()
+	{
+		clientThread.invoke(() ->
+		{
+			final InventorySetup active = panel.getCurrentSelectedSetup();
+			final Map<String, Object> data = active == null ? Map.of() : Map.of(API_DATA_SETUP, active.getName());
+			eventBus.post(new PluginMessage(API_NAMESPACE, API_MSG_ACTIVE_SETUP_CHANGED, data));
+		});
+	}
+
 	public void handleMessage(final PluginMessage message)
 	{
 		if (!API_NAMESPACE.equals(message.getNamespace()))
 		{
 			return;
 		}
-		if (API_MSG_SETUPS_CHANGED.equals(message.getName()))
+		if (API_MSG_SETUPS_CHANGED.equals(message.getName()) || API_MSG_ACTIVE_SETUP_CHANGED.equals(message.getName()))
 		{
 			// Our own outgoing broadcast.
 			return;
@@ -107,6 +136,11 @@ public class InventorySetupsPluginMessageHandler
 			case API_MSG_CLEAR:
 			{
 				handleClear(message);
+				break;
+			}
+			case API_MSG_GET_ACTIVE_SETUP_CONTENTS:
+			{
+				handleGetActiveSetupContents(message);
 				break;
 			}
 			default:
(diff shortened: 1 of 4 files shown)


src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -188,6 +222,52 @@ private void handleClear(final PluginMessage message)
...
+		{
+			final InventorySetup setup = panel.getCurrentSelectedSetup();
+			final boolean hasActiveSetup = setup != null && setup.isFilterBank() && plugin.isFilteringAllowed();
+			message.getData().put(API_DATA_HAS_ACTIVE_SETUP, hasActiveSetup);
+			if (!hasActiveSetup)
+			{
+				return;
+			}
+
+			final Collection<Integer> equipmentItemIds = asIntegerCollection(equipmentObj);
```
dillydill123: You should be able to replace this with `setup.getSetupItems()` to simplify

Robbejj: Using `getSetupItems()` would cause other issues there.
It would merge all of the items into a single list leaving the api consumer pretty clueluess about which item is which.
I guess it could be fine for inventory and equipment (as they are fixed length), but the additional filtered items would probably include rune pouch, bolt pouch and quiver and we would have no way to tell where the additional items end.

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -80,13 +97,26 @@ public void broadcastSetupsChanged()
...
 
+	// Broadcasts the active setup's name, or clears it. Called on selection change and on every content
+	// edit of the setup that's still active. Deliberately not deduped like broadcastSetupsChanged() -
+	// content edits don't change the name, but callers still need to know to re-read the contents.
+	public void broadcastActiveSetupChanged()
+	{
+		clientThread.invoke(() ->
+		{
+			final InventorySetup active = panel.getCurrentSelectedSetup();
+			final Map<String, Object> data = active == null ? Map.of() : Map.of(API_DATA_SETUP, active.getName());
```
dillydill123: nit: Perhaps `API_DATA_ACTIVE_SETUP` instead of `API_DATA_SETUP` for this key?

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -40,9 +43,23 @@ public class InventorySetupsPluginMessageHandler
...
 	public static final String API_MSG_CLEAR = "clear";
+	// out: broadcast when the active setup changes - on opening, closing and editing
+	// of the setup that's still active (add/remove an item, toggle fuzzy, etc.)
+	// data["setup"] = the active setup's name; the key is absent when setup is closed.
+	public static final String API_MSG_ACTIVE_SETUP_CHANGED = "active-setup-changed";
+	// in: get the active setup's contents by slot, e.g. for a plugin that wants to mirror its layout elsewhere.
+	// Put mutable Collection<Integer> under "equipmentItemIds" (EquipmentInventorySlot order,
+	// size 14), "inventoryItemIds" (size 28), and "additionalItemIds" (no position semantics). Posting is synchronous.
+	// data["activeSetup"] is set to the active setup's name (String) when one is active and
+	// bank filtering is allowed; the key is absent otherwise.
```
dillydill123: nit: Add a note in this comment that `-1` indicates no item in that slot

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -188,6 +223,49 @@ private void handleClear(final PluginMessage message)
...
+		if (!(equipmentObj instanceof Collection) || !(inventoryObj instanceof Collection) || !(additionalObj instanceof Collection))
+		{
+			log.warn("Ignoring {} message without the expected mutable collections", API_MSG_GET_ACTIVE_SETUP_CONTENTS);
+			return;
+		}
+
+		clientThread.invoke(() ->
+		{
+			final InventorySetup setup = panel.getCurrentSelectedSetup();
+			final boolean hasActiveSetup = setup != null && setup.isFilterBank() && plugin.isFilteringAllowed();
```
dillydill123, line 241: nice!

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -188,6 +223,49 @@ private void handleClear(final PluginMessage message)
...
+			if (!hasActiveSetup)
+			{
+				return;
+			}
+			message.getData().put(API_DATA_ACTIVE_SETUP, setup.getName());
+
+			final Collection<Integer> equipmentItemIds = asIntegerCollection(equipmentObj);
```
dillydill123: Could you combine the `asIntegerCollection` and `stream(...)`? Something like the following (did not test)

```
private static void addItemIds(
    final Object obj,
    final Collection<InventorySetupsItem> items)
{
    final Collection<Integer> ids = (Collection<Integer>) obj;

    items.stream()
        .map(InventorySetupsItem::getId)
        .forEach(ids::add);
}
```

Then just one function call for each to reuse this?

```
addItemIds(inventoryObj, setup.getInventory());
addItemIds(inventoryObj, setup.getEquipment());
addItemIds(additionalObj, setup.getAdditionalFilteredItems().values());
```

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -188,6 +223,49 @@ private void handleClear(final PluginMessage message)
...
+			setup.getInventory().stream()
+				.map(item -> item.getId())
+				.forEach(inventoryItemIds::add);
+			setup.getAdditionalFilteredItems().values().stream()
+				.map(InventorySetupsItem::getId)
+				.forEach(additionalItemIds::add);
+		});
+	}
+
+	@SuppressWarnings("unchecked")
```
dillydill123: nit: Scope to the cast of the local variable

src/main/java/inventorysetups/ui/InventorySetupsPluginPanel.java
```diff
@@ -754,6 +754,7 @@ public void setCurrentInventorySetup(final InventorySetup inventorySetup, boolea
 	{
 		overviewPanelScrollPosition = contentWrapperPane.getVerticalScrollBar().getValue();
 		currentSelectedSetup = inventorySetup;
+		plugin.broadcastActiveSetupChanged();
```
dillydill123: nit: Move to near line 782 with the rest of the `plugin` calls at the end for consistency

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -80,13 +98,26 @@ public void broadcastSetupsChanged()
...
 
+	// Broadcasts the active setup's name, or clears it. Called on selection change and on every content
+	// edit of the setup that's still active. Deliberately not deduped like broadcastSetupsChanged() -
+	// content edits don't change the name, but callers still need to know to re-read the contents.
+	public void broadcastActiveSetupChanged()
+	{
+		clientThread.invoke(() ->
+		{
+			final InventorySetup currentSetup = panel.getCurrentSelectedSetup();
+			final Map<String, Object> data = currentSetup == null ? Map.of() : Map.of(API_DATA_ACTIVE_SETUP, currentSetup.getName());
```
dillydill123: Output `""` if no active setup instead of making the key completely absent,

src/main/java/inventorysetups/InventorySetupsPluginMessageHandler.java
```diff
@@ -40,9 +43,24 @@ public class InventorySetupsPluginMessageHandler
...
+	// out: broadcast when the active setup changes - on opening, closing and editing
+	// of the setup that's still active (add/remove an item, toggle fuzzy, etc.)
+	// data["activeSetup"] = the active setup's name; the key is absent when setup is closed.
+	public static final String API_MSG_ACTIVE_SETUP_CHANGED = "active-setup-changed";
+	// in: get the active setup's contents by slot, e.g. for a plugin that wants to mirror its layout elsewhere.
+	// Put mutable Collection<Integer> under "equipmentItemIds" (EquipmentInventorySlot order, size 14),
+	// "inventoryItemIds" (size 28) and "additionalItemIds" (no position semantics).
+	// Empty slots are included with id -1 to avoid breaking proper order. Posting is synchronous.
+	// data["activeSetup"] is set to the active setup's name (String) when one
+	// is active and bank filtering is allowed; the key is absent otherwise.
```
dillydill123: Update comment here after

dillydill123 (approved):

